PR #203, opened 2026-06-12
Fix #189: live .deslop.toml exclude edits re-evaluate the corpus

### LINKED ISSUE #189: Live .deslop.toml `exclude` edits are not reactive — UI keeps stale offenders (rerender-only)
## Summary

Live `.deslop.toml` **`exclude`** edits are **not reactive**. Adding an `exclude`
pattern while the live server is running (VS Code extension / LSP / MCP) does **not**
drop the now-excluded files from the corpus. The reactive loop
(watcher → scheduler → session → broadcast → UI) fires, the config is reloaded, but
the session only **re-renders** — it never re-evaluates the existing corpus against the
new exclusion. Stale offenders keep showing in the UI until a full restart.

Only `report_hide` (render-time) works live today. `exclude` (discovery-time) was never
wired into the live reload. This breaks the core "Deslop.live (reactive) means the whole
loop" promise for the single most common config edit.

## Repro

1. Open a workspace in VS Code with the Deslop extension live (e.g. the
   `flutter-wonderous-app` sample). `TOP OFFENDERS` shows
   `lib/l10n/app_localizations_zh.dart` cluster `c37741f` **rank #1, 71 copies**.
2. Add `<root>/.deslop.toml`:
   ```toml
   [defaults]
   exclude = ["lib/l10n/**"]
   ```
3. Save. The watcher forwards the config path (it bypasses the extension filter via
   `watched_config_paths`), `reload_exclusion()` runs, and the log shows
   `deslop_toml_reloaded mode=rerender-only`.

**Expected:** the `lib/l10n/**` files are dropped from the live corpus; `TOP OFFENDERS`
refreshes and l10n disappears (mirroring the CLI).

**Actual:** the UI is unchanged — l10n is still rank #1. A **cold CLI run** with the
identical `.deslop.toml` correctly drops them (**250 → 96 groups, ~1009 KB → 107 KB**),
proving the config is valid and only the *live* path is broken.

## Root cause

`AnalysisSession::apply_changes` — `crates/deslop-core/src/live/session.rs:359`:

- Detects `config_changed` and calls `pipeline.reload_exclusion()` (session.rs:370–380) —
  so the new `ExclusionConfig` **is** loaded into the pipeline.
- Then **filters the config path out of the changed set** (session.rs:382–386), so
  `run_pipeline` is called with an **empty** `source_changed`.
- Exclusion is only ever re-checked against a file inside
  `PipelineSession::apply_one_change` —
  `crates/deslop-core/src/pipeline/session/change.rs:38–41` (`is_excluded` → `drop_path`)
  — which only runs for paths **in the changed set**.

Net: already-discovered files that a newly-added `exclude` pattern now matches are
**never re-evaluated and never `drop_path`'d**. The docstring/`mode=rerender-only` log
confirm this was implemented for `report_hide` only ([LIVE-CONFIG-LIVE], #139).

## Proposed fix direction

On a config change, after `reload_exclusion()`, re-evaluate the **existing corpus**
against the reloaded exclusion — `drop_path` every currently-live path the new config now
excludes, and (for completeness) re-discover to pick up paths a removed pattern now
re-includes — instead of `rerender-only`. Then bump the generation and broadcast so the
UI refreshes.

Fix forward via the mandatory fix-bug TDD process. A coarse E2E test fits the existing
`--rerun-*` CLI harness (which already simulates watcher-driven `update_files` between
generations): assert that introducing an `exclude` pattern between generation N and N+1
removes the matching files from the `ReportDelta` / rendered report.

## Severity — showstopper

The headline "reactive" guarantee is silently broken for the most common config edit.
Users edit `.deslop.toml`, get no error, and keep staring at stale results with no signal
that a restart is required.

## Related

- Likely-related second defect found while diagnosing: in
  `crates/deslop-core/src/config.rs:519`, `matches()` calls `matched(path, false)` (file
  paths, no parent-dir walk), so a standard gitignore **directory** pattern like
  `lib/l10n/` silently excludes nothing — you must write `lib/l10n/**`. Worth a separate
  issue.

Wamwea:
## TLDR

Live `.deslop.toml` `exclude` edits now re-evaluate the whole corpus instead of only re-rendering, so newly-excluded files drop out of the live report and files re-admitted by a removed pattern are re-discovered — without a restart. Closes #189.

## Details

**Root cause.** `AnalysisSession::apply_changes` (`crates/deslop-core/src/live/session.rs`) reloaded the exclusion config when `.deslop.toml` changed, but then filtered the config path out of the changed set, so `PipelineSession::update_files` ran with an empty list and only re-rendered (`mode=rerender-only`). Exclusion was only ever re-checked inside `apply_one_change`, which runs solely for paths in the changed set — already-discovered files matching a new `exclude` pattern were never dropped, and files re-admitted by a removed pattern were never re-discovered.

**Fix — `crates/deslop-core` (`[LIVE-CONFIG-LIVE]`):**

- `PipelineSession::update_files` (`pipeline/session/mod.rs`) now detects watched config paths (`<root>/.deslop.toml` or the explicit `--config` override) in the changed set and calls the new `refresh_exclusion` before the per-file pass.
- `refresh_exclusion` (`pipeline/session/change.rs`) reloads the config, then `drop_newly_excluded` sweeps every live path against the reloaded `exclude` patterns and `drop_path`s the matches, and `apply_newly_included` re-runs `discover_files` and applies every path the new config re-admits that the corpus does not yet contain. A malformed config is logged and the prior exclusion kept, preserving the "a typo never bricks the daemon" guarantee. Structured `tracing` reports the dropped/added counts.
- `AnalysisSession::apply_changes` (`live/session.rs`) no longer pre-filters config paths or reloads the config itself — it forwards the changed set verbatim, so the CLI rerun harness and the live daemon exercise the identical code path. The generation bump and broadcast were already in place, so the reactive loop (watcher → scheduler → session → broadcast → UI) now delivers correct data end-to-end.
- `reload_exclusion` is demoted from `pub` to private — `update_files` is now its only caller.

**Deduplication.** The `watched_config_paths` / `is_config_path` / `canonicalise_or_clone` helpers move from `live/session.rs` into `config.rs` (public), and `deslop-lsp/src/file_watch.rs` deletes its near-duplicate `watched_config_paths` in favour of the shared one. One definition now serves the pipeline session, the live session, and the LSP watcher bypass list.

**Breaking changes:** `PipelineSession::reload_exclusion` is no longer public (it had no external callers besides the live session, which no longer needs it). No CLI flag, report format, or spec ID changes.

Not addressed here (tracked separately per the issue): gitignore-style **directory** patterns (`lib/l10n/`) still match nothing because `is_excluded` matches file paths without a parent-dir walk — consistent between CLI and live, worth its own issue.

Two new black-box E2E tests in `crates/deslop/tests/rerun.rs` drive the CLI binary against the `csharp-small` fixture (Alpha.cs + Beta.cs form a clone cluster) using the `--rerun-*` harness, which replays watcher-style changes through `PipelineSession::update_files` between generation 0 and 1 — the same entry point the live daemon uses:

- `issue_189_new_exclude_pattern_drops_existing_corpus_files` — `--rerun-add` lands a `.deslop.toml` with `exclude = ["**/Beta.cs"]` at the scan root between generations and replays the config path. Asserts the emitted `report.delta.json` has `from_generation` 0 → `to_generation` 1 with `clusters_removed` non-empty, and that the generation-1 `report.json` no longer mentions `Beta.cs`. Before the fix this produced an empty delta (the config path was a silent no-op).
- `issue_189_removed_exclude_pattern_rediscovers_files` — the reverse direction, via the explicit `--config` override path: generation 0 excludes `Beta.cs` (no clusters), the rerun swaps in a config with `exclude = []`. Asserts `clusters_added` is non-empty and `Beta.cs` appears in the generation-1 report, proving re-discovery of re-admitted files. Also failed with an empty delta before the fix.

Both tests were written first and confirmed failing against the unfixed code (empty `clusters_removed` / `clusters_added`), then confirmed passing after the fix. The existing seven `rerun.rs` tests still pass, covering the no-op, re-parse, unsupported-path, add, remove, and pre-loaded-exclusion branches around the new code.

### DIFF
--- a/crates/deslop-core/src/config.rs
+++ b/crates/deslop-core/src/config.rs
@@ -696,6 +696,38 @@ fn matches(matcher: &Gitignore, path: &Path) -> bool {
     matcher.matched(path, false).is_ignore()
 }
 
+/// Builds the canonical set of config paths that should trigger a live
+/// exclusion reload — `<root>/.deslop.toml` plus the explicit override
+/// (if any) ([LIVE-CONFIG-LIVE], #139).
+#[must_use]
+pub fn watched_config_paths(root: &Path, override_path: Option<&Path>) -> Vec<PathBuf> {
+    let default = root.join(DEFAULT_CONFIG_FILENAME);
+    let mut paths = vec![canonicalise_or_clone(&default)];
+    if let Some(explicit) = override_path {
+        paths.push(canonicalise_or_clone(explicit));
+    }
+    paths
+}
+
+/// Returns `true` when `candidate` matches one of the watched config
+/// paths in either canonical or as-given form.
+#[must_use]
+pub fn is_config_path(candidate: &Path, watched: &[PathBuf]) -> bool {
+    if watched.iter().any(|watched_path| watched_path == candidate) {
+        return true;
+    }
+    let canonical = canonicalise_or_clone(candidate);
+    watched
+        .iter()
+        .any(|watched_path| watched_path == &canonical)
+}
+
+/// Canonicalises `path` when possible; otherwise returns a clone so
+/// non-existent override paths still compare predictably.
+fn canonicalise_or_clone(path: &Path) -> PathBuf {
+    std::fs::canonicalize(path).unwrap_or_else(|_| path.to_path_buf())
+}
+
 /// Compiles a list of glob patterns into an [`ignore::gitignore::Gitignore`]
 /// matcher. The builder is rooted at the scan root when known so user
 /// patterns are scan-root-relative — `subdir/**` matches
--- a/crates/deslop-core/src/live/session.rs
+++ b/crates/deslop-core/src/live/session.rs
@@ -343,13 +343,12 @@ impl AnalysisSession {
     /// empty delta — the queued paths are replayed inside
     /// [`Self::install_pipeline`].
     ///
-    /// When any entry in `changed` is a watched config file
-    /// (`<root>/.deslop.toml` or the explicit override), the live
-    /// exclusion config is reloaded before re-rendering so a
-    /// `report_hide` edit hides the matching cluster on the next
-    /// generation ([LIVE-CONFIG-LIVE], #139). Source-file entries are
-    /// re-fingerprinted as before; config-file entries are dropped
-    /// from the per-file pass because they are not source.
+    /// Watched config files (`<root>/.deslop.toml` or the explicit
+    /// override) in `changed` are detected inside
+    /// [`crate::PipelineSession::update_files`], which reloads the
+    /// exclusion config and re-evaluates the existing corpus against
+    /// it — dropping newly-excluded files and re-discovering newly
+    /// re-included ones ([LIVE-CONFIG-LIVE], #139, #189).
     ///
     /// # Errors
     ///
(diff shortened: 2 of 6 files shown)


Wamwea:
Heads-up: #205 contains this exact changeset plus the #204 watcher-extension fix and Shipwright script fixes. If #205 merges first, this PR can be closed.

PR #205, opened 2026-06-12
Derive LSP watched extensions from the parser registry; make .deslop.toml exclude edits reactive

### LINKED ISSUE #189: Live .deslop.toml `exclude` edits are not reactive — UI keeps stale offenders (rerender-only)

### LINKED ISSUE #204: LSP watcher ignores .dart files — reactive loop dead for pure-disk Dart edits
## Symptom

Pure filesystem writes to `.dart` files (the primary path for agent-driven edits, which write to disk rather than through LSP `didChange`) never reach the live scheduler. The reactive loop (watcher → scheduler → session → broadcast → UI) is dead for Dart source edits until the server restarts.

Editor-driven edits partially mask the bug because `did_change`/`did_open` also feed `apply_changed_paths`, and MCP reads self-heal via `refresh_if_stale` — but a pure external save of a Dart file is invisible.

## Root cause

`crates/deslop-lsp/src/file_watch.rs` hardcoded a parallel extension list that drifted when Dart support landed:

```rust
const WATCHED_EXTENSIONS: &[&str] = &["cs", "rs", "py"];
```

The analysis engine's own extension filter is derived from the `LanguageParser` registry (`default_parsers()` → `build_extension_map`), so the initial full scan analyses Dart fine — only the incremental watcher path dropped `.dart` events ([LIVE-WATCHER], [PIPELINE-LANG-TRAIT]).

## Fix

Derive the watched-extension list from the registered parser set (new `deslop_core::pipeline::watched_source_extensions()`, beside `language_ids()`), deleting the hardcoded constant — so TypeScript/Go can never reintroduce the drift.

Regression E2E test: `report_changed_fires_for_external_save_of_dart_file` in `crates/deslop-lsp/tests/notifications.rs` — spawns the real `deslop-lsp` binary on the `dart-small` fixture, writes `beta.dart` on disk with no LSP notification, and requires a `deslop/reportChanged` push reporting the removed clone cluster. Times out before the fix; passes after.

Related but distinct: #170 / #198 (MCP language enum omitting `dart`).

Wamwea:
## TLDR

Fixes two live-loop reactivity bugs: the LSP filesystem watcher now derives its extension filter from the `LanguageParser` registry so pure-disk `.dart` edits drive the reactive loop (Closes #204), and live `.deslop.toml` `exclude` edits now re-evaluate the whole corpus instead of only re-rendering (Closes #189).

## Details

**Watcher extension drift — Closes #204 ([LIVE-WATCHER], [PIPELINE-LANG-TRAIT])**

- `crates/deslop-lsp/src/file_watch.rs` hardcoded `WATCHED_EXTENSIONS = &["cs", "rs", "py"]` — a parallel copy of the language set that drifted when Dart landed. Pure filesystem writes to `.dart` files (the primary path for agent-driven edits) never reached the scheduler, so the reactive loop (watcher → scheduler → session → broadcast → UI) was dead for Dart until restart.
- New `watched_source_extensions()` in `crates/deslop-core/src/pipeline/corpus.rs` (exported from `deslop_core::pipeline` beside `language_ids()`) flat-maps `LanguageParser::file_extensions()` over `default_parsers()`. The watcher now consumes it; the hardcoded constant is deleted. The parser registry is the single source of truth, so future languages (TypeScript, Go) cannot reintroduce the drift.

**Live `.deslop.toml` exclude edits — Closes #189 ([LIVE-CONFIG-LIVE])**

- Previously a config-file change only reloaded the exclusion config and re-rendered, so the UI kept stale offenders. `PipelineSession::update_files` (`crates/deslop-core/src/pipeline/session/mod.rs`) now detects watched config paths in the changed set and calls the new `refresh_exclusion` (`crates/deslop-core/src/pipeline/session/change.rs`), which reloads the config, drops corpus files the new `exclude` patterns match (`drop_newly_excluded`), and re-discovers files a removed pattern re-admits (`apply_newly_included`). A malformed config is logged and the prior exclusion kept, so a typo never bricks the daemon.
- `watched_config_paths` / `is_config_path` / path canonicalisation moved from `crates/deslop-core/src/live/session.rs` into `crates/deslop-core/src/config.rs` as public helpers; `AnalysisSession::apply_changed_paths` no longer does its own config-path filtering, and `reload_exclusion` is no longer public (it is now an internal step of `refresh_exclusion`). `crates/deslop-lsp/src/file_watch.rs` consumes `config::watched_config_paths` instead of a private duplicate.

**Shipwright verifier scripts — path-with-spaces fix**

- `scripts/test-release-workflow-contract.mjs`, `scripts/test-release-version-stamping.mjs`, and `scripts/test-verifiers.mjs` derived the repo root via `new URL("..", import.meta.url).pathname`, which leaves percent-encoding in place (`Coding%20Tools`) and aborts `make ci` with `ENOENT` on any checkout whose path contains a space. They now use `fileURLToPath` from `node:url`; unused `resolve` imports removed.

**Docs**

- `CLAUDE.md`: the agent directive for `find-similar` is expanded into the "LAW: call `find-similar` BEFORE you author new code" section with fused-score thresholds, tool-selection guidance, and hard rules.

**Breaking change note:** `deslop_core::PipelineSession::reload_exclusion` is no longer `pub`; config reload is now driven through `update_files`. No CLI flag, report format, or spec ID changed.

- `report_changed_fires_for_external_save_of_dart_file` (`crates/deslop-lsp/tests/notifications.rs`, [LIVE-WATCHER]) — black-box E2E against the real `deslop-lsp` binary on the `dart-small` fixture. Asserts the Dart clone cluster exists at startup, then writes `beta.dart` on disk **without any LSP notification** and requires a `deslop/reportChanged` push whose `summary.clusters_removed > 0`. Timed out before the fix (watcher dropped the `.dart` event); passes after.
- `issue_189_new_exclude_pattern_drops_existing_corpus_files` (`crates/deslop/tests/rerun.rs`, [LIVE-CONFIG-LIVE]) — replays a `.deslop.toml` landing at the watched path via `--rerun-add`; asserts the generation-1 delta removes the Alpha/Beta clone cluster and the rendered report no longer mentions the excluded `Beta.cs`.
- `issue_189_removed_exclude_pattern_rediscovers_files` (`crates/deslop/tests/rerun.rs`, [LIVE-CONFIG-LIVE]) — the reverse direction through the explicit `--config` override: loosening the exclude re-discovers `Beta.cs`, the delta reports `clusters_added > 0`, and the report includes the re-included file.
- The Shipwright proof suites themselves (`3 release workflow contract tests passed`, `3 release version stamping tests passed`, `31 verifier proof tests passed`) now run to completion from a space-containing checkout, where they previously aborted before executing a single test.
- Full `make ci` (lint + fail-fast test suite + build + deployment-verify) passes locally with coverage above every threshold in `coverage-thresholds.json` (workspace 94.8%; thresholds unchanged).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,11 +17,52 @@
 
 **The point of Deslop is to prevent duplication, not just scrub it afterwards.** Post-hoc deduplication is what every static analyzer already does. Deslop's edge is being **live in the agent's inner loop** so a duplicate never lands in the first place.
 
-**Directive for AI agents (Claude Code, Cursor, Copilot, Codex, Continue):**
+### LAW: call `find-similar` BEFORE you author new code
 
-> **Before writing any new function, class, helper, or test setup, call the `find-similar` MCP tool first.** If the proposed pattern already exists with high similarity, reuse the canonical implementation instead of authoring a new copy.
+Before you write any new function, method, class, helper, fixture, test setup,
+parser branch, error type, route handler, view model, or any other code unit
+larger than a few lines, you MUST call the `find-similar` MCP tool with the
+proposed code (or its byte range, if it already exists in a draft buffer) and
+inspect the response.
 
-This applies during **code authoring**. When fixing existing duplicates, use `top-offenders` and `cluster-by-id`; when writing fresh code, use `find-similar`. Both modes use the same live server — the call you make encodes the intent.
+- If the response shows a cluster with **`signals.fused ≥ 0.85`** *or* the
+  bucket is `identical` / `nearly_identical`, do NOT write the new copy.
+  A `structural_only` match means only the code shape lines up (no token
+  or semantic evidence) — read the match before deciding; it is often
+  sibling boilerplate rather than a reusable implementation.
+  Reuse the canonical occurrence the tool returns. Extract a helper if needed.
+- If the response is empty or the closest match is structurally distant
+  (`signals.fused < 0.6`), proceed with authoring.
+- If the response is borderline (`0.6 ≤ fused < 0.85`), read the canonical
+  occurrence and decide whether the new code is genuinely different. Bias
+  toward reuse.
+
+`find-similar` accepts either:
+- a **byte range** in an existing file (`path`, `start_byte`, `end_byte`), or
+- a **snippet** (`snippet`, `language`) — useful when you have a draft in your
+  scratchpad that has not been written to disk yet.
+
+### When to use the OTHER Deslop tools instead
+
+- **Fixing existing duplicates** (refactor / dedup work) → start with
+  `top-offenders`, then `cluster-by-id` for the cluster you'll merge. Don't
+  use `find-similar` for this — it answers a different question.
+- **Investigating a specific file** → `report-for-file`.
+- **Investigating a specific block before refactor** → `report-for-range`.
+- **Schema reference for the JSON shapes** → call `schema-doc` *once* per
+  session. Don't bundle it into every response.
+
+### Hard rules
+
+- Do NOT skip `find-similar` because "this looks new." Most clones are
+  introduced by people who genuinely thought the code was new. The tool exists
+  to override that intuition with evidence.
+- Do NOT silence Deslop warnings by widening thresholds, marking clusters
+  hidden, or splitting the code into trivially different shapes. If the tool
+  flags something, treat it as a real signal until you've shown otherwise.
+- Do NOT call `find-similar` after writing a duplicate "just to confirm." That
+  is cure, not prevention. Call it BEFORE the keystroke that creates the
+  duplicate.
 
 The MCP tool description leads with prevention ("Call BEFORE writing new code…"). Reports surface `find-similar` as the prevention path in their `action_hints`. This document and the user-facing `AGENTS.md` recipe teach this so agents pick it up without re-reading docstrings.
 
--- a/crates/deslop-core/src/config.rs
+++ b/crates/deslop-core/src/config.rs
@@ -696,6 +696,38 @@ fn matches(matcher: &Gitignore, path: &Path) -> bool {
     matcher.matched(path, false).is_ignore()
 }
 
+/// Builds the canonical set of config paths that should trigger a live
+/// exclusion reload — `<root>/.deslop.toml` plus the explicit override
+/// (if any) ([LIVE-CONFIG-LIVE], #139).
+#[must_use]
+pub fn watched_config_paths(root: &Path, override_path: Option<&Path>) -> Vec<PathBuf> {
+    let default = root.join(DEFAULT_CONFIG_FILENAME);
+    let mut paths = vec![canonicalise_or_clone(&default)];
+    if let Some(explicit) = override_path {
+        paths.push(canonicalise_or_clone(explicit));
+    }
+    paths
+}
+
+/// Returns `true` when `candidate` matches one of the watched config
+/// paths in either canonical or as-given form.
+#[must_use]
+pub fn is_config_path(candidate: &Path, watched: &[PathBuf]) -> bool {
+    if watched.iter().any(|watched_path| watched_path == candidate) {
+        return true;
+    }
+    let canonical = canonicalise_or_clone(candidate);
+    watched
+        .iter()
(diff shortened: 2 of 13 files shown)


Wamwea:
Note for review: this PR is a strict superset of #203 — both branches were cut from the same working tree, so the #189 changeset (config.rs, live/session.rs, pipeline/session/*, file_watch.rs, rerun.rs) is byte-identical in both. Merging this PR makes #203 redundant (close it); alternatively merge #203 first and this one will rebase down to the #204 watcher fix + Shipwright script fixes.

PR #265, opened 2026-07-03
Add PHP language parser support

dseguy:
## TLDR
Adds PHP as a first-class parsed language, including the tree-sitter grammar pin, default parser registration, normalisation rules, fixtures, and E2E coverage.

## Details
- Adds the exact `tree-sitter-php = "=0.24.2"` workspace dependency in `Cargo.toml`, wires it into `crates/deslop-core/Cargo.toml`, and updates `Cargo.lock`.
- Extends `.github/workflows/ci.yml` so the grammar-pin check covers `tree-sitter-php` alongside the existing tree-sitter grammars.
- Adds `crates/deslop-core/src/lang/php.rs`, exporting `PhpParser` from `crates/deslop-core/src/lang/mod.rs` and registering it in `default_parsers()` in `crates/deslop-core/src/pipeline/corpus.rs`.
- Implements PHP normalisation for `[PARSE-PHP-NORMALIZE]`: `name` collapses to `__ident__`, PHP literal node kinds collapse to `__literal__`, comments/doc-comments are dropped, and all other named node kinds pass through via the shared parser plumbing used by `[PIPELINE-LANG-TRAIT]` and `[PIPELINE-NORMALIZE-AST]`.
- Extends `function_kinds()` in `crates/deslop-core/src/cluster_filters/mod.rs` so PHP `function_definition` and `method_declaration` bodies participate in the signature-only and polymorphic-signature filters.
- Adds PHP fixture coverage under `crates/deslop/tests/fixtures/`: `php-small` renamed-clone fixtures, `ast-golden-php` normalisation golden files, and `php-enums` fixtures for enum/match syntax.
- No public API, CLI flag, report format, or existing spec ID is removed. Behaviour changes intentionally because `.php` files are now included in default analysis and can produce findings where they were previously skipped.

- `debug_ast_dump_matches_committed_golden_php` validates the committed PHP AST golden, proving PHP comments/doc-comments are dropped, identifiers and literals are collapsed, and core class/method/control-flow/cast nodes stay structurally stable under `[PARSE-PHP-NORMALIZE]`.
- `detects_type2_clone_in_php_fixture` runs the CLI against the `php-small` fixture repo and asserts both PHP files are analysed, both renamed functions appear in the JSON report, and the clone has `"structural": 1.0`.
- Local `make ci` passed after the change, including Rust coverage thresholds, VSIX coverage, packaged VSIX E2E, webview coverage, and the standalone HTML report Playwright gate. Coverage thresholds were not ratcheted in this PR.


MelbourneDeveloper:
Thankyou @dseguy!

The best way to work through the ci issues before submitting is run `make ci`. That will tell you what will get stuck in the pipeline. You can also run /ci-prep with your AI agent and the agent should be able to fix it up for you.

image

Issue #264, opened 2026-07-03
find-similar snippet query misses a tracked structural cluster (recall false negative)


Wamwea:
## Symptom

A `find-similar` snippet query returned `{"clusters": [], "total_occurrences": 0}` for a Rust snippet that is near-verbatim identical to both occurrences of tracked cluster `a59932844b88648e` (the duplicated `ReportFixture` test scaffold).

- Corpus occurrences: `crates/deslop-core/tests/issue_121_pytest_fixture_boilerplate.rs:82-172` and `crates/deslop-core/tests/issue_98_99_108_120_122_thresholds.rs:171-261` (structural=1.00, 457-node subtree).
- `report-for-file` on either file returns the cluster correctly — the corpus/cluster state is fine; only the snippet-input path missed.

## Query that missed

`find-similar` with `language: "rust"` and a snippet containing the `ReportFixture` struct declaration plus its `new()` and `member()` methods copied verbatim from issue_121 (with `cluster()`/`render()` omitted and one string literal changed `"python"` → `"csharp"`). Expected: the `a59932844b88648e` cluster (or at least a `structural_only` match well above `fused 0.6`). Got: empty.

## Why it matters

`find-similar` is the prevention gate agents must call before authoring code (CLAUDE.md rule zero). A snippet that is a strict subset of a tracked 457-node clone returning *zero* signal defeats the gate exactly where it should fire. Suspect the snippet path only matches whole-subtree fingerprints at matching roots and cannot match a snippet that is a subset/superset of an indexed subtree.

## Environment

- deslop-lsp 0.0.0-dev built from main @ 862ea36, workspace = the Deslop repo itself, `min_nodes=30`, embeddings off.
- Discovered while working GH #239.

MelbourneDeveloper:
**Cluster: find-similar snippet-mode recall** — #309 #264

Root cause: **#309**. Snippet mode returns `{"clusters": [], "total_occurrences": 0}` for a byte-for-byte verbatim copy of corpus code that is not already clustered — the first-copy prevention gap. #264 is the same path missing a snippet that is near-verbatim to both occurrences of an *already tracked* cluster, so it is the narrower case of the same query path. In both, `report-for-file` over the same code returns the cluster correctly, which places the defect in the snippet-input path and not in corpus or cluster state.

Fix order: #309 → #264.

Blocking edge recorded: #264 ← #309.

[...]

PR #261, opened 2026-07-02
Group by duplicate type everywhere: HTML report bucket expanders (#257), facet surfaces (#195), flat bucket type-mode in the panel (#258)

Wamwea:
Fixes #257. Fixes #258.

## TLDR

Group duplicates by type everywhere: the HTML report gets per-bucket collapsible expanders with CSS-only facets (#257), the remaining facet surfaces from #195 land, and the VS Code panel's "Group by Type" mode now groups **flat by clone bucket** — all Identical clusters in one group — instead of by finding category (#258).

### HTML report bucket groups + CSS-only facets (#257) — `[FACET-HTML]`

- **Bucket-group expanders** (`crates/deslop-core/src/render/html_buckets.rs`, new): cluster cards group into one `<details class="bucket-group kind-…">` per bucket present, worst-first across and within groups; the worst group starts `open`, the rest collapsed. Summaries carry the shared plain title + live count (e.g. *Identical code — 2 group(s)*). Applies inside the flat "Duplicate groups" section and inside each per-language section alike.
- **CSS-only facet filter**: a "Show:" row of checkbox-driven chips, one per bucket present (rendered only when ≥2 buckets exist). Unchecking a chip hides that bucket's groups via a sibling selector. **No JS** — the artifact must stay inert on `file://` and inside the VSIX's script-disabled report tab (`enableScripts: false` in `clients/vscode/src/webview/htmlReport.ts`).
- **Shared vocabulary, zero drift**: every label, class suffix, and selector derives from the canonical bucket registry (`bucket_labels`/`classify` in `buckets.rs`, mirrored by `bucketLabels`/`resolveBucket` in `clients/vscode/src/types/report.ts`) per `[CLONE-BUCKETS-DUAL-LABEL]`. Facet CSS is generated from the registry at render time — nothing hand-listed.
- `group_clusters_by_language` now delegates to the same generic first-seen grouper as the bucket groups (extract-and-reuse; `html.rs` net-shrank).

### Remaining facet surfaces (#195) — `[FACET-TOP-OFFENDERS-FILTER]`, `[FACET-REPORT-WEBVIEW]`, `[FACET-CLI]`

The bucket/category facet filter across the Top Offenders tree, report webview, status-bar count, and CLI summary breakdown, all funnelling through the one shared `applyFacetFilter` slice.

### Panel type mode groups flat by bucket (#258) — `[FACET-GROUP-BY-TYPE]`

[...]

PR #266, opened 2026-07-03
Parse each C# file once per render — fix the per-member full-file re-parse hanging large C# corpora (#239)

### LINKED ISSUE #239: Showstopper: VSCode panel stuck on "Analysing…" forever — LSP logs `render complete` (5534 clusters) but webview never renders
### Severity
**Showstopper.** The Deslop sidebar (TOP OFFENDERS + SESSION panes) is permanently pinned to the "Analysing…" spinner. The extension is unusable as a result — no offenders, no clusters, no session data ever appear.

### Environment
- Extension: `nimblesite.deslop-vscode-0.0.0-dev`
- Version: `0.0.0-dev`
- Platform: `darwin` / `arm64`
- LSP binary: `bundled` `deslop-lsp` (`0.0.0-dev`)
- Workspace: 452 files (`csharp: 428, python: 1, rust: 23`), `min_nodes=30`, `embedding_mode=off`

### Symptom
Both **TOP OFFENDERS** and **SESSION** panes show `⟳ ⋮ Analysing…` indefinitely. **FOCUSED FILE** shows `No active editor` (expected — no editor open). The spinner never resolves even though the backend log shows the analysis pipeline **completed successfully**.

### Smoking gun — the LSP finished, the UI did not
The backend log shows the full pipeline ran to completion in **~21 seconds**, yet the webview stays in the loading state. The render-complete signal from the LSP is evidently **not reaching / not being honored by the webview**, OR the webview is not subscribing to the completion notification.

```
22:16:17.307  deslop-lsp args parsed  workspace_root=.../DataProvider min_nodes=30 worker_threads=0 embedding_mode="off"
22:16:17.307  run_stdio booting backend  min_nodes=30
22:16:17.307  pipeline session initialising  incremental=true supported_extensions=["rs","cs","dart","py"]
22:16:17.355  file discovery complete  total_files=452 by_language={"csharp":428,"python":1,"rust":23}
22:16:19.541  fingerprint corpus built  files_processed=452 fingerprints=303917 cache_hits=443 cache_misses=9
22:16:37.365  pair survival outcome  total=471539 survived=170585 dropped_below_fused=275106 ...
22:16:38.170  ranked clusters built  total=5534 input_total=19936 dropped_below_min_members=400 largest_weight=2618373.34 mean_weight=10033.00
22:16:38.170  render complete  ranked_clusters=5534 fingerprints=303917   <-- PIPELINE DONE
```

`render complete  ranked_clusters=5534` was logged at `22:16:38`. The screenshot of the perpetually-spinning UI was taken well after that. **The pipeline is done; the webview just never transitions out of the loading state.**

### Expected
After `render complete`, the TOP OFFENDERS pane should populate with the ranked clusters (5534 were produced) and SESSION should show summary stats. The spinner should clear.

### Actual
Spinner spins forever in both panes. No data ever shown.

### Likely root cause (for triage)
The render-complete handoff from LSP → extension host → webview is dropping the final result on a **large result set** (5534 ranked clusters / 303917 fingerprints). Candidates:
1. The webview's "analysing" state is only cleared by a notification the host fails to forward (or forwards before the webview has registered its listener — a startup race).
2. The result payload is too large to post to the webview in one message and the send silently fails (no error surfaced to the UI — note there is **zero diagnostic** in the panel, same silent-failure class as #173).
3. The webview re-mounts (e.g. on focus/visibility change) and loses the already-delivered result, falling back to the loading state with no re-request.

This is the same **silent-no-diagnostic** failure family as #173 ("blank panel on large/active repos"). The UI gives the user no error, no retry, no indication anything went wrong — it just spins.

### Asks
1. **Surface a diagnostic** in the panel when the render-complete payload is not applied (never leave a bare spinner).
2. Add a **timeout / watchdog** on the "Analysing…" state that flips to an error + retry affordance if no result lands within N seconds after the LSP reports `render complete`.
3. Confirm the webview **re-requests** the latest snapshot on (re)mount instead of waiting passively.
4. Verify large payloads (5k+ clusters) survive the host→webview `postMessage` boundary (chunk if needed).

### Repro
1. Open a repo of ~450 source files (mixed C#/Rust/Python) in VSCode with the Deslop extension.
2. Open the Deslop sidebar.
3. Observe TOP OFFENDERS + SESSION stuck on "Analysing…" despite the LSP log emitting `render complete`.

Wamwea:
## TLDR
Fixes the GH #239 showstopper (VSCode panel pinned on "Analysing…" forever) by making `render_report`'s C#-only byte-equivalence fallback reuse the shared per-render `ParseCache` instead of re-parsing the entire source file once per cluster member.

## Details
**Root cause.** `csharp_methods_in_range` in `crates/deslop-core/src/report_render.rs` constructed a fresh tree-sitter `Parser` and re-parsed the full file for every member of every C# cluster. It runs whenever a C# cluster's members are not byte-equivalent (i.e. nearly every cluster) via `materialise_cluster` → `cluster_to_report` → `report_bucket_kind` → `source_slices_are_equivalent_for_language`. On a large C# corpus (the reporter's workspace: 428 `.cs` files, 5,534 ranked clusters) this cost minutes-to-hours of single-threaded CPU inside `LspBackend` construction — before the LSP could answer `initialize` — so the editor spinner never resolved and the seed cache was never written (every restart re-scanned and re-hung).

**The fix** (`deslop-core`):
- `crates/deslop-core/src/report_render.rs` — `csharp_methods_in_range` now walks a cached CST (`&tree_sitter::Tree`) instead of parsing; the per-render `ParseCache` is threaded `cluster_to_report` → `report_bucket_kind` → `source_slices_are_equivalent_for_language` → `csharp_method_declarations_are_equivalent`. Each file is parsed at most once per report render, matching the `[CLONE-NOISE-REPARSE-CACHE]` contract the noise filters already follow. The `[CLONE-BUCKETS-IDENTICAL]` equivalence-upgrade semantics are unchanged (a file whose parse fails still refuses to prove equivalence).
- `crates/deslop-core/src/cluster_filters/snippets.rs` — `ParseCache::tree_for` widened from private to `pub(crate)` so the renderer can share the cache.
- `crates/deslop-core/src/report.rs` — `materialise_cluster` passes the `ParseCache` it already owned into `cluster_to_report`.

**Test scaffolding dedup (required by the new test):** the `ReportFixture` helper was duplicated byte-for-byte across `tests/issue_121_pytest_fixture_boilerplate.rs` and `tests/issue_98_99_108_120_122_thresholds.rs` (tracked by Deslop's own report as cluster `a59932844b88648e`). Rather than adding a third copy it now lives once in `tests/common/mod.rs`, parameterised by language, with `file()` / `fingerprint_at()` so one file can carry several cluster members at distinct byte ranges. Both existing suites were updated to consume it; their assertions are untouched. Net diff across the PR is negative (−41 lines).

**No breaking changes.** All touched functions are `pub(crate)` or private; no wire model, report format, CLI flag, or spec ID changed. Measured effect: on a synthetic 100-file C# Type-2 corpus, the CLI's render phase (`render complete` → `bucket distribution`) drops from 24.8s to 0.79s (~31×); the reporter's corpus stops hanging entirely.

Related: GH #264 (filed during this work — `find-similar` recall false negative observed while querying the prevention gate).

Closes #239.

- **New:** `crates/deslop-core/tests/issue_239_csharp_reparse.rs::csharp_equivalence_fallback_scales_with_files_not_members_issue_239` drives the production `render_report` path over two generated Type-2 C# files (192 renamed methods each) twice: a two-member baseline cluster and a 384-member stress cluster. It first asserts the stress cluster renders correctly (visible, bucket `nearly_identical`, all 384 occurrences), then asserts the stress render stays within `max(20 × baseline, 250ms)` — parse work must scale with the number of files, not members. The bound is a self-calibrating ratio, so it holds across debug/release/coverage hosts. **Before the fix it fails at 12,734ms against a 2,800ms budget (91× the 140ms baseline); after the fix it passes with an order-of-magnitude margin.**
- **Unchanged-behaviour proof for the dedup:** `issue_121_pytest_fixture_boilerplate` and `issue_98_99_108_120_122_thresholds` re-run against the consolidated `ReportFixture` with identical assertions (cluster hiding, bucket routing, `clusters_hidden` counts), demonstrating the shared fixture reproduces the exact render semantics of the deleted copies.
- Full `make test` passes with coverage enforced: `deslop-core` 94.4%, `deslop` 97.1%, `deslop-lsp` 95.0%, `deslop-mcp` 97.5% — all at or above their `coverage-thresholds.json` gates.


MelbourneDeveloper:
Superseded by #268, which consolidates this C# render-cache fix with #261 and the CI-prep fixes on one branch. Closing this original so review and CI stay on the combined PR.

Issue #309, opened 2026-07-26
find-similar returns empty for a verbatim copy of unclustered corpus code (first-copy prevention gap)


lhaig:
## Symptom

`find-similar` (snippet mode, `language: "rust"`) returns `{"below_min_nodes": false, "clusters": [], "total_occurrences": 0}` for a snippet that is a **byte-for-byte verbatim copy** of code already in the analysed corpus: the entire `impl LanguageParser for FSharpParser` block, `crates/deslop-core/src/lang/fsharp.rs:64-85`.

Control checks in the same session:
- `report-for-file` on `crates/deslop-core/src/lang/fsharp.rs` returned 0 clusters (the block is unique in the corpus today).
- `report-for-file` on `crates/deslop-core/src/lang/javascript.rs` correctly returned tracked structural cluster `8cadabc104716594` (sibling parser impl boilerplate), so corpus/cluster state is healthy.

## Why it matters

The tool's own description says: "Call BEFORE writing new code to PREVENT duplication". The highest-value moment for that gate is when an agent is about to write the **first** copy of code that currently exists exactly once. Because the snippet path appears to search only existing *clusters* (not the corpus subtree fingerprint index), unique corpus code is invisible to the gate, and the first copy always lands undetected. Rule zero's decision table ("fused < 0.6 -> author it") then green-lights the exact clone the tool exists to prevent.

Related but distinct from #264: that report is about snippet recall against an already-*tracked* cluster; this one is about *unclustered* (single-occurrence) corpus code, which can never match under cluster-only search regardless of the #264 subset/root-alignment issue.

## Reproduction

1. Workspace = the Deslop repo itself; start deslop-lsp, connect deslop-mcp (`--root` at repo root).
2. Copy `crates/deslop-core/src/lang/fsharp.rs` lines 64-85 verbatim (the whole `impl LanguageParser for FSharpParser` block, well above min-nodes).
3. Call `find-similar` with `{snippet: <that text>, language: "rust"}`.
4. Expected: an `identical`-bucket (or at minimum high-fused) match pointing at `fsharp.rs:64`. Got: empty result.

## Environment

- deslop 0.28.0 / deslop-lsp 0.28.0 / deslop-mcp 0.28.0 (Homebrew), macOS arm64.
- Discovered while running the rule-zero gate for new-language work (Go parser), where the planned code mirrors `fsharp.rs` and the gate returned no signal.

MelbourneDeveloper:
[...]

Issue #69, opened 2026-05-02
False positive: ABC method signature duplication across interface implementations


MelbourneDeveloper:
## Summary

Deslop reports cluster **7a7740f11f48d9f3** as 7 identical copies of a 37-node subtree — but all 7 are implementations of the same abstract method on an ABC (`AgentWorkspaceHost`).

## Reproduction

In `NimblesiteAgenticPlatform`, `AgentWorkspaceHost` defines:

```python
@abstractmethod
async def tool_call(
    self,
    *,
    instance: AgentWorkspaceContainer,
    name: str,
    arguments: dict[str, Any],
    trace_id: str | None = None,
) -> Result[ToolCallResult, AgentWorkspaceError]:
    ...
```

Concrete subclasses (`FlyAgentWorkspaceHost`, `DockerAgentWorkspaceHost`, `MockAgentWorkspaceHost`, test stubs) each implement this method. The 37-node subtree is the **method signature**, which MUST be identical across all implementations — that is the contract.

## Expected behaviour

Method signatures implementing the same abstract interface should not be flagged as duplicate code. The fix would be to detect when occurrences share a common base class/ABC and suppress the cluster, or at minimum add an `interpretation` note that the duplicate is an interface implementation pattern.

## Impact

This generates false noise on every Python project using ABCs. Closes discussion with `[structural=1.00, token_jaccard=1.00]` scores.

MelbourneDeveloper:
Verified fixed on current `main`. Polymorphic method overrides / ABC-style interface implementations that share one method name across files but carry distinct bodies are suppressed by `is_polymorphic_signature_cluster` (`crates/deslop-core/src/cluster_filters/mod.rs`), wired language-agnostically into `is_noise_pattern`. A byte-identical copy still surfaces (carve-out `enclosing_function_bodies_differ`).

Proof (run just now, passes): `cargo test -p deslop --test dart_signatures dart_polymorphic_override_signatures_are_suppressed` → ok (the cluster is actively hidden while a genuine clone still surfaces). The suppression logic is the same across languages; a Python regression test wired to the existing `python-issue-69-abstract-method` fixture would be a nice follow-up but the behaviour is fixed and covered.

Closing as fixed.

abdushakoor12:
Tracked as #373 — the over-correction described above now has its own open issue with the reproduction, the six-variant probe table, and acceptance criteria for both sides (this issue's ABC case must stay suppressed).

PR #310, opened 2026-07-26
feat: add Go language support

lhaig:
## TLDR

Adds Go as a first-class analysis language (`[LANG-CAND-GO]`, P-LANG-3): `tree-sitter-go` pinned, `GoParser` implemented, and the full per-language E2E proof (Type-2, Type-3, dissimilar guard, AST golden) wired through every surface.

## Details

- **New `crates/deslop-core/src/lang/go.rs`** implementing `LanguageParser` (`[PIPELINE-LANG-TRAIT]`) over `tree-sitter-go = "=0.25.0"` — a new workspace dependency, pinned exactly in `Cargo.toml` and added to the ci.yml grammar-pin verify loop. Normalisation follows `[LANG-CAND-GO]` with documented extensions: `blank_identifier`/`label_name` also collapse to `__ident__`; the string `*_content` nodes, `escape_sequence`, and `iota` collapse to `__literal__`; `composite_literal`/`literal_value`/`literal_element` and `func_literal` closures stay structural so shape remains the clone signal.
- **Registry wiring**: `pipeline::corpus::default_parsers`, `cluster_filters::mod::function_kinds` (`function_declaration`/`method_declaration`/`func_literal` — all expose a `body` field, enabling the `[CLONE-NOISE-SIGNATURE-ONLY]`/#154 and #69 filters), `cluster_filters::snippets::grammar_for`, and `render::html::language_display_name`. Discovery, the live extension map, and the MCP `language` enum derive from the parser registry and pick Go up with no further wiring.
- **Boilerplate classification**: `package_clause` and `import_declaration` are import/prologue carriers (`[PIPELINE-BOILERPLATE]`) — the Go analogue of C# `using` directives.
- **Test-helper extraction in `crates/deslop/tests/cli/detection.rs`**: the live `find-similar` gate flagged the planned Go Type-2 test as the eighth copy of a tracked 30-node cluster, so the per-language assert bodies were extracted into shared helpers (`assert_type2_report`, `report_clusters`, `cluster_file_basenames`, `require_cluster_spanning`, `assert_type3_signals`, `assert_every_cluster_single_file`). The six existing Type-2 tests, the F# Type-3 test, and the three dissimilar guards now use them — the file shrinks 482 → 471 lines while gaining three Go tests, with assertion-for-assertion parity.
- **`crates/deslop-core/tests/live.rs`**: the unsupported-language error-path probe switches from `go` to `kotlin` (`[LANG-CAND-KOTLIN]` stays deferred on grammar blockers), preserving the `LiveError::UnsupportedLanguage` assertion now that `go` parses.
- **VS Code**: `onLanguage:go` + `workspaceContains:**/*.go` activation events.
- **JetBrains**: `DeslopSupportedFiles` and the LSP4IJ `fileNamePatternMapping` gain `go` — plus the `php`/`fs`/`fsx` entries the PHP/F# slices never added. The file's "keep in lockstep with the languages deslop-lsp parses" contract had silently drifted (site docs already claimed both languages worked). `DeslopSupportedFilesTest` updated: the shipping-extension list covers all fourteen extensions and `java` replaces `go` as the dormant probe. Note: no JVM was available locally, so this test relies on the CI job.
- **Docs**: README, site EN + zh (`index.njk`, `how-it-works.md`, `ai-integration.md`), `docs/specs/SPEC.md` and `docs/specs/pipeline.md` language sets, `docs/plans/LANG-ROADMAP.md` (P-LANG-3 moved to Shipped with the normalisation deltas recorded), `docs/plans/PLAN.md` (Java is the next language slice), and the stale CLAUDE.md "languages today" list.
- Filed #309 during the work: `find-similar` returns empty for a verbatim copy of unclustered corpus code (first-copy prevention gap), discovered while running the rule-zero gate for this change.

**No breaking changes**: no public API, CLI flag, report format, or spec ID changed; the MCP `language` enum gains `go` additively.

- `detection::detects_type2_clone_in_go_fixture` — the `go-small` pair (`alpha.go`/`beta.go`: same accumulate loop, every identifier renamed, integer literals changed) must analyse 2 files and cluster at `structural: 1.0`, proving Go identifier/literal collapse delivers Type-2 rename invariance.
- `detection::detects_type3_clone_in_go_fixture` — `go-type3` (`delta.go` runs two accumulator updates per loop iteration, `epsilon.go` one) must produce a cross-file cluster spanning both files at structural = 1.0 with ≥ 2 occurrences and a `token_jaccard` signal, while the signature-only sibling match (`func f(_ int) int` with differing bodies) is suppressed — proving both the near-miss path and the `[CLONE-NOISE-SIGNATURE-ONLY]`/#154 filter are live for Go.
- `detection::dissimilar_go_functions_across_files_stay_in_separate_clusters` — zero-false-positive guard: a map-building range loop vs an early-return `if` chain must never pair; every cluster's occurrences stay single-file.
- `cache_and_debug::debug_ast_dump_matches_committed_golden_go` — `--debug-ast` on `Sample.go` (labels, `iota`, rune/imaginary/raw-string literals, qualified `strings.Builder` type, blank identifier, expression switch, composite literal) must match the committed golden byte-for-byte, so any grammar bump or `normalise_kind` edit that would silently invalidate user caches trips the test.
- `live_service_round_trip_covers_the_query_surface` — still proves the `UnsupportedLanguage` error path, now via `kotlin`.
- The ten converted tests (six Type-2, F# Type-3, three dissimilar guards) assert exactly what they asserted before through the shared helpers, so the dedup refactor is behaviour-preserving by construction.
- Coverage floors in `coverage-thresholds.json` are unchanged and met on the full local `make ci`: deslop-core 94.7%, deslop 97.2%, deslop-lsp 94.4%, deslop-mcp 97.6%, VSIX 94.3%, webview 96.3%; Playwright HTML-report suite 3/3.


MelbourneDeveloper:
Awesome. Thankyou @lhaig

Issue #339, opened 2026-08-09
F# clone pairs score token_jaccard from the issue-86 fallback signature — byte-offset luck, not token evidence


abdushakoor12:
## Summary

For F# module-level clone pairs, the token stream backing the MinHash signature is empty/unresolvable, so `build_signatures_with_languages` falls back to the fingerprint-scoped signature from #86 (`crates/deslop-core/src/pipeline/signatures.rs`, `fallback_signature`: blake3 over `hash + byte_range`). `token_jaccard` for such pairs is then an artifact of whether the two fingerprints' byte ranges coincide — not token evidence.

## Repro (deslop @ current main, `--min-nodes 20 --embeddings off --no-incremental`)

Two files, identical 10-line function, only the module name differs:

```fsharp
module ParseHelpers      // parse_b.fs: module ParseHelpersB

let accumulate (values: int list) (floor: int) =
    let mutable total = 0
    for value in values do
        if value > floor then
            total <- total + value * 2
        else
            total <- total - 1
    total
```

| variant | occurrence ranges | token_jaccard | bucket |
|---|---|---|---|
| byte-identical pair | (7,245) / (7,245) | 1.00 | identical |
| same-length rename (`ParseHelperX`) | (7,245) / (7,245) | 1.00 | nearly_identical |
| one-char-longer rename (`ParseHelpersB`) | (7,245) / (7,246) | **0.00** | **structural_only** |

The token layer is documented rename-invariant (normalised node kinds; `assert_bucketed_clone` in `crates/deslop/tests/common/mod.rs` asserts exactly this). A length-changing rename must not flip `token_jaccard` from 1.0 to 0.0.

## Impact

- Renamed F# Type-2 clones are misrouted to `structural_only` (demoted) with `token_jaccard=0.00`.
- Where ranges happen to coincide, the 1.00 reading is a fallback-signature collision, not evidence — it silently inflates confidence for F#.

## Suspected fix area

Why the F# token stream is empty for these fingerprints: `tokens::token_stream_for_fingerprint*` / `locate` against the F# normalised tree (module-level binding subtrees, possibly synthetic sibling windows). Found while writing regression fixtures for #331/#336.

abdushakoor12:
## Resolution: symptom fixed on `fix/331-336-fused-signal-independence`, pinned by regression test

[...]

Issue #336, opened 2026-08-09
False positive: numeric array literals rank #1 on dotnet/fsharp — data-table classification is Dart-only


MelbourneDeveloper:
## Summary

Scanning `dotnet/fsharp` @ `v15.2.302` (`ba48c78076ef15cdfef89d0001c538f3f130421b`), the **top-ranked cluster in the entire report** is an integer array literal, classified `category: "logic"`.

`CloneCategory::data` exists (#190) but its only classifier is `crates/deslop-core/src/cluster_filters/dart_data_table.rs`. Every non-Dart language therefore reports data tables at full `logic` weight.

Found during accuracy-audit corpus testing.

## Repro

```bash
git clone --depth 1 --branch v15.2.302 https://github.com/dotnet/fsharp.git
deslop fsharp --embeddings off --no-incremental
```

## The #1 cluster

`size=3544`, `weight=145263`, `bucket=nearly_identical`, `category=logic`, `signals.fused=1.0`

`tests/FSharp.Compiler.LanguageServer.Tests/Protocol.fs:180`:

```fsharp
[| 0; 0; 0; 1; 0; 0; 0; 3; 15; 0; 0; 4; 1; 17; 0; 0; 4; 1; 19; 0 |]
```

**3,544 occurrences** of a 67-byte numeric literal, ranked as the single worst duplication in the F# compiler.

## Ranks 2-4 are the same class

| # | size | weight | what it is |
|---|---|---|---|
| 0 | 3544 | 145263 | integer array literal |
| 1 | 291 | 138620 | `buildtools/fsyacc/fsyacclex.fs` — **generated** lexer tables (`# 1 "fsyacclex.fsl"`) |
| 2 | 68 | 106865 | `ComparersRegression.fs` — generated test data, numeric array literals |
| 3 | 108 | 97263 | same file, same shape |

**Zero** of the top 50 clusters carry `category: "data"`.

## Impact

F# reports **42.8% duplication** across 6,233 files / 934,677 LOC — the highest of the nine languages measured, and the top of that report is data literals and generated code rather than anything a developer could act on.

Generated-file suppression also appears not to fire on `fsyacclex.fs`, which carries a generator preamble.

## Expected

`CloneCategory` classification should be language-agnostic, or at minimum implemented per language, so numeric/data tables are demoted under the existing `[ranking] data_clones` policy instead of outranking real logic clones.

abdushakoor12:
[...]

Issue #331, opened 2026-08-09
False positive: Flutter widget declaration boilerplate ranks #1 on flutter/flutter


MelbourneDeveloper:
## Summary

Scanning `flutter/flutter` @ `3.38.9` (`67323de285b00232883f53b84095eb72be97d35c`), the **top-ranked cluster in the entire report** is the framework-mandated `StatefulWidget` declaration. It is not actionable duplication — the Flutter framework requires every `StatefulWidget` to declare its own `createState()`, so it cannot be extracted or merged.

Found during accuracy-audit corpus testing.

## Repro

```bash
git clone --depth 1 --branch 3.38.9 https://github.com/flutter/flutter.git
deslop flutter --embeddings off --no-incremental
```

## The #1 cluster

`size=453`, `weight=14464`, `bucket=nearly_identical`, `signals={structural:1.0, token_jaccard:1.0, fused:1.0}`

```dart
// examples/layers/services/lifecycle.dart:7-12
class LifecycleWatcher extends StatefulWidget {
  const LifecycleWatcher({super.key});
  @override
  State<LifecycleWatcher> createState() => _LifecycleWatcherState();
}

// examples/layers/widgets/spinning_square.dart:7-12
class SpinningSquare extends StatefulWidget {
  const SpinningSquare({super.key});
  @override
  State<SpinningSquare> createState() => _SpinningSquareState();
}
```

Identifier normalisation collapses the class names, so `token_jaccard` reaches 1.0 and the cluster is bucketed `nearly_identical` rather than routed to `structural_only`.

## Not an isolated case

The next-ranked clusters are the same class of framework scaffolding:

| # | size | weight | shape |
|---|---|---|---|
| 0 | 453 | 14464 | `StatefulWidget` + `createState()` |
| 1 | 182 | 13032 | `StatelessWidget` + `build()` returning `MaterialApp`/`Scaffold` |
| 5 | 192 | 8022 | `StatelessWidget` + `build()` returning `MaterialApp` |

Report-wide: 21,305 clusters, 43,856 hidden, 22.5% duplication on 2,370,585 LOC.

## Why this matters

Ranking is the user-visible product and it is "worst offenders first". A user opening this report is told the most valuable cleanup in Flutter is boilerplate they are forbidden from removing. Every genuine duplicate is pushed below it.

## Analysis

[...]

Issue #301, opened 2026-07-25
duplication_percent is nondeterministic across runs on an unchanged tree (0.191pp spread over 9 runs) — makes --fail-over CI gates flaky


MelbourneDeveloper:
## Summary

`deslop .` returns a **different `metrics.duplication_percent` on every run over a byte-identical tree**. Across 9 consecutive runs (no edits, no `--incremental`, no cache) the same corpus measured anywhere from **9.353% to 9.544%** — a spread of **0.191 percentage points**.

Because `--fail-over` / `[threshold] max_duplication_percent` gates CI on exactly this number, and CI runs the tool **once**, a repo sitting anywhere within ~0.2% of its threshold gets a **randomly red build**. The threshold has to carry ~0.2% of dead slack purely to absorb the noise, which directly defeats the ratchet-down workflow the tool is designed around.

## Environment

- deslop **0.26.0** (`deslop-0.26.0-macos-arm64.tar.gz`, official release asset)
- macOS 25.5.0, arm64
- Corpus: Dart monorepo, 68,621 analysed LOC, ~370 clusters
- Embeddings **off** (default) — so this is not embedding-provider nondeterminism

## Reproduction

```bash
for i in $(seq 1 9); do
  deslop . --output "run$i" --nohtml --notext --no-fail-over >/dev/null 2>&1
  python3 -c "import json;m=json.load(open('run$i.json'))['metrics'];\
print('%.3f  dup=%d loc=%d clusters=%d'%(m['duplication_percent'],m['duplicated_loc'],m['analysed_loc'],m['clusters_total']))"
done
```

## Observed

`analysed_loc` is constant at 68621 every run (parsing is stable). `duplicated_loc` is not:

| run | duplication_percent | duplicated_loc | clusters |
|----:|--------------------:|---------------:|---------:|
| 1 | 9.461 | 6492 | 371 |
| 2 | 9.544 | 6549 | 370 |
| 3 | 9.536 | 6544 | 372 |
| 4 | 9.423 | 6466 | 370 |
| 5 | 9.423 | 6466 | 372 |
| 6 | 9.366 | 6427 | 368 |
| 7 | 9.353 | 6418 | 368 |
| 8 | 9.383 | 6455 | 369 |
| 9 | 9.474 | 6483 | 370 |

**min 9.353 · max 9.544 · spread 0.191 · mean 9.440 · stdev 0.069**
`duplicated_loc` swings by **131 lines** run-to-run on identical bytes.

[...]

PR #341, opened 2026-08-09
Content-gate fused confidence so shape-only families stop outranking genuine clones

abdushakoor12:
## TLDR

Give the fused confidence an independent evidence source — raw-content agreement measured at the normalisation-collapsed leaves — so shape-only families (framework boilerplate, data tables, renamed scaffolding) stop saturating to `fused = 1.0` and outranking genuine clones, validated against the pinned real-repository corpus.

Related to these issues

#331

#336

#339

#283

#284

#285

#103

#79 

#71

## Details

**Root cause attacked** (`[FUSION-CONTENT-GATE]`, new spec section in `docs/specs/fusion.md`): `structural` and `token_jaccard` are both computed from the identifier/literal-stripped representation, so on any shape match they agree by construction and sum-then-clamp fusion pinned `fused` at 1.0 — the mechanism behind eight open false-positive issues and both accuracy showstoppers.

**New: `crates/deslop-core/src/content.rs`.** A per-render pass (`attach_content_agreement`, wired in `pipeline/session/render.rs`) walks each cluster member's normalised tree and hashes the raw source bytes at every collapsed leaf (`__ident__`/`__literal__` — the walk lives in `tokens.rs::collapsed_leaves`, sharing `resolve_range_nodes` with token extraction so both signals see the same code). Per cluster it stores on `cluster::Cluster`:

- `content_agreement` — mean positional match vs the canonical member for shape-identical members, content-key set Jaccard for shape-mismatched (LSH-paired) members, with a proportional verbatim guard: ≥ half the members byte-identical vouches full agreement (the #104 mixed-cluster case), a stray verbatim pair inside a mega-family does not.
- `literal_fraction` — share of collapsed leaves that are literal positions (min 8 literals).

**Fusion and routing** (`buckets.rs`, `report_render.rs`):

- `content_gated_signals`: rendered `fused = max(embedding_cos, max(structural, token_jaccard) × content_agreement)` for clusters with saturating shape evidence (`structural ≥ 0.99` **or** `token_jaccard ≥ 0.95` — the LSH kind-stream is a shape echo too). Byte-equivalence-proven `identical` clusters keep 1.0; the GH #232-style token correction applies only under the Merkle-equality argument (`structural ≥ 0.99`).
- `route_shape_identical`: three zones — content < 0.7 (`CONTENT_SUPPORT_FLOOR`, the `[TECH-TOKEN-SOURCERERCC]` cutoff) demotes to the `structural_only`/hidden-scaffolding tier; ≥ 0.85 (`CONTENT_PROMOTE_FLOOR`) rescues fallback-signature artifacts to `nearly_identical` (small spreads only); between, the legacy verdict stands (preserves the human-validated #197 REST-family suppression at 0.72–0.80).
- `[RANK-STRUCTURAL-ONLY]` weight demotion now exempts `category=data` clusters so the `[ranking] data_clones` policy (`report.rs::structural_only_multiplier`) keeps owning table weight — `data_clone_weight = 1.0` can still restore a table.

**Language-agnostic data tables** (`[CLONE-NOISE-LITERAL-TABLE]`, new spec section in `docs/specs/exclusion.md`; `cluster_filters/mod.rs`): a cluster whose canonical member is ≥ 0.8 literal positions with ≥ 2 byte-differing members classifies `category: "data"` in every language — previously Dart-only (#336's headline complaint). Literal-dominated shape-only families route to the surfaced `structural_only` tier instead of the hidden one, so the data policy governs their visibility.

**Flutter widget scaffolds** (`[CLONE-NOISE-DART-WIDGET-SCAFFOLD]`; `cluster_filters/dart.rs::is_dart_widget_scaffold_cluster`): clusters covering only whole classes extending `StatelessWidget`/`StatefulWidget`/`State<T>` (plus the mandated `runApp` launcher) are hidden as framework-mandated noise — template-stamped docs examples are honest near-copies no content measure can demote. Containment rules in both directions keep body-level clones surfacing as their own subtree clusters.

**Corpus ratchet** (`corpus/known-failures.json`): `flutter/boilerplate_rank` (#331) and `fsharp/data_table_rank` (#336) deleted after passing against the pinned real repositories; remaining entries map to #166 (memory) and #301 (determinism), untouched here.

**Test-support and fixtures:** F# corpus builders and rank/category helpers consolidated in `crates/deslop/tests/common/mod.rs`; 17 rename-showcase fixtures softened from maximal renames (which the honest engine now routes `structural_only`) to partial renames so they stay truthful `nearly_identical` seeds — each keeps its feature-specific variation. One expectation change: `typescript_features.rs` `ts-type-keyword` now expects `nearly_identical` (its pair shares every name at content 0.94; the old `structural_only` expectation encoded the #339 token-fallback artifact).

**Breaking-behaviour note (no wire change):** the JSON report schema is unchanged, but rendered `fused`, `token_jaccard`, `bucket`, and `category` values shift for shape-only clusters — consumers keying on `fused ≥ 0.85` now get the documented act-now semantics instead of a constant 1.0. `report_render.rs` remains over the 500-line rule (pre-existing, 647 lines) — tracked as debt.

**New E2E regression binaries** (black-box CLI runs over fixture corpora, per the repo's testing rules):

- `issue_331_336_shape_only_saturation.rs` — six distinct `StatefulWidget` declarations and four same-length/different-value F# arrays must not reach the act-now fused line or an act-now bucket, and must rank below genuine byte-identical clones whose `identical` bucket, `token_jaccard = 1.0`, and `fused ≥ 0.85` are asserted as recall guards; a third test pins that template-stamped `ExampleApp` scaffolds (same class name, high content) surface in **no** visible cluster while the genuine clone survives. Both saturation tests failed on `main` with `bucket=nearly_identical, fused=1` before the fix.
- `fsharp_issue_336_data_table_category.rs` — the F# table family carries `category: "data"`, stays visible-but-demoted under the default policy, disappears under `data_clones = "ignore"` (counted in `clusters_hidden`), out-ranks the logic clone under `data_clone_weight = 1.0`, and a byte-identical copied table stays `logic` at `identical` (the #190 escape hatch). The default and keep-mode tests failed on `main` (category was `logic`, the knob inert for F#).
- `fsharp_issue_339_token_fallback_rename.rs` — a module rename one character longer shifts every byte offset; the test pins `token_jaccard = 1.0`, `nearly_identical`, and act-now fused (on `main` the pair read `token_jaccard = 0.00` and misrouted to `structural_only`).

**Real-repository corpus gate** (`make test-corpus`, `[CORPUS-PRECISION]`/`[CORPUS-RECALL]`): on pinned `flutter/flutter` 3.38.9 (6,231 files / 2.37M LOC) no `StatefulWidget`/`StatelessWidget` shape appears in the top-5 and all three hand-verified byte-identical recall entries are still reported; on pinned `dotnet/fsharp` v15.2.302 the top numeric-table clusters all carry `category: data`. Both checks previously failed and their baseline entries are deleted, so any regression now fails scheduled CI outright.

**Updated suites:** `dart_issue_197_single_file_structural_only.rs` (REST families stay hidden — proves the middle band preserves human-validated suppression), `python_issue_104_module_preamble.rs::verbatim_copy_among_lookalikes_still_surfaces` (proportional verbatim guard keeps the mixed cluster visible), `issue_190_data_table_demote.rs::keep_mode_restores_data_table_to_the_top` (data policy overrides the structural-only demotion), and the js/ts signature/normalization/feature suites re-validated over the softened fixtures — `token_jaccard = 1.0` rename-invariance assertions unchanged throughout.

Full `make ci` green locally: lint (deny-wall clippy), fail-fast suite, coverage above every crate threshold in `coverage-thresholds.json`.


crates/deslop/tests/fixtures/javascript-type3/delta.js
MelbourneDeveloper, line 1: There are a bunch of test fixtures here that have just been changed...  Is there a real reason why they were changed? Couldn't we keep the old fixtures AND add new ones for new tests? Just want to make sure there's a solid reason for just replacing tests

corpus/known-failures.json
```diff
@@ -17,18 +15,22 @@
...
+    ],
+    "nest": [
+      "determinism"
+    ],
+    "jellyfin": [
+      "determinism"
+    ]
   },
-
   "_issues": {
```
MelbourneDeveloper, line 32: Just checking why these changed... Is there a reason for this?

MelbourneDeveloper (approved):
This is good as long we can explain why so many tests were changed instead of adding new tests that test for different things

crates/deslop/tests/fixtures/javascript-type3/delta.js
abdushakoor12: the fix deliberately changes what those fixtures' code means to the engine. And the fixtures, not the tests, were encoding the old bug.

**The mechanism**: most of those fixtures were maximal renames — every identifier changed between the two files (accumulate/values/running/step → aggregate/items/accumulator/cursor). They were written when the token layer couldn't distinguish a rename from a copy, so a fully-renamed pair still scored token_jaccard=1.0 and bucketed nearly_identical. That is exactly the false-positive signature of #331 — the Flutter widget cluster is nothing but a fully-renamed family. After the fix, a pair with every name different is honestly structural_only: shape is the only evidence.

That left two options for the tests that used these fixtures:

1. Keep the fixtures, change the assertions — expect structural_only and drop the token_jaccard ≈ 1.0 checks. That weakens assertions, which this repo's rules prohibit, and it would have quietly re-labelled feature-coverage tests as bucket-policy tests.
2. Keep every assertion, adjust the fixtures so they're honest examples of the bucket they claim to demonstrate — a copy that renames the function and a parameter but keeps its body names, which is also what real copy-paste looks like.

Option 2 was taken.

Issue #352, opened 2026-08-11
cpu_throttle_knob LSP test is timing-dependent and carries a lint suppression


Wamwea:
## TL;DR

One of our own tests is unreliable: it waits a fixed 1.5 seconds for a program to write a startup line to its log, then kills it and checks the log. On a slow moment the line hasn't been written yet, so the test fails with an empty log. It also contains a linter suppression, which the project forbids. Both should be fixed.

## Details

The test starts the Deslop language server, sleeps for exactly 1.5 seconds, kills it, and then checks that a startup line ("which CPU setting was used") made it into the error output. Whether the line arrives in time is down to luck — the project's own rules say tests must not depend on timing like this, and must not contain linter suppressions; this test breaks both.

It failed this way on macOS on 2026-08-12: the log was empty after the wait, while the neighbouring test in the same file passed. CI can't confirm either way — the CI pipeline hasn't run since 2026-05-11.

The fix direction is already clear: instead of sleeping and hoping, read the output line by line until the startup line shows up (with the normal test timeout as the bound), then kill. Found while checking the fix for #347; left out of that change to keep it small.

## Steps To Reproduce

Fails intermittently (observed on macOS arm64, 2026-08-12):

1. Run the test:

   ```
   cargo test -p deslop-lsp --test cpu_throttle_knob lsp_startup_log_records_the_worker_threads_knob
   ```

2. If it passes, repeat under load until it fails.
3. The failure message ends with an empty log: `stderr was:` — nothing after it.

## Details (for AI)

```shell
# evidence
test: crates/deslop-lsp/tests/cpu_throttle_knob.rs :: lsp_startup_log_records_the_worker_threads_knob
observed: macOS arm64 2026-08-12 — empty stderr after the 1.5s window; sibling test in the same binary passes
ci status: main ci.yml not run since 2026-05-11 (latest run cancelled)

# rule violations (CLAUDE.md)
1. thread::sleep(Duration::from_millis(1500)) — "Deterministic. No sleep, no timing dependencies, no random state."
   mechanism: spawns deslop-lsp, sleeps 1.5s, kills the child, asserts the startup tracing line reached stderr — pass depends on whether the tracing buffer flushed before the kill
2. #[allow(unused_mut)] — "No linter suppressions."

# fix direction
deterministic wait: read stderr line-by-line until the startup line appears (bounded by the harness timeout), then kill
found while verifying the fix for #347; not addressed there to keep that diff scoped
```

[...]

Issue #342, opened 2026-08-10
Built-in exclude components match ancestors of the scan root — a repo under a folder named `dist`/`build`/`target` is silently analysed as zero files


Wamwea:
## Summary

`built_in_excluded` in `crates/deslop-core/src/config.rs` tests **every component of the absolute discovered path** against `BUILTIN_EXCLUDE_COMPONENTS`. That includes components *above* the scan root, so pointing deslop at a repository that merely *lives under* an ancestor directory named `dist`, `build`, `target`, `vendor`, `node_modules`, … excludes every single file:

- `files_analysed: 0`
- `clusters: []`
- exit code: success

A total, silent false negative: the user scans their repo, gets a clean empty report, and believes their code has no duplication. Found while investigating #298 — adding `out` there (which I support) widens this surface further (Chromium-style workspaces, `~/out/…`, any checkout under a `build` folder).

## Repro (current `main`)

```
mkdir -p /tmp/dist/innocent-repo/src
# write two .ts files containing a genuine cross-file clone into src/
deslop /tmp/dist/innocent-repo --no-incremental --min-nodes 8 --embeddings off
# → files_analysed: 0, no clusters
```

Move the same repo to `/tmp/innocent-repo` and rerun: `files_analysed: 2`, the clone pair is reported. The only difference is the name of a directory the user never asked deslop to reason about.

## Expected

Built-in component exclusion should only consider components **below the scan root**. The user explicitly chose the root; ancestors are not part of the analysed corpus. The codebase already encodes exactly this principle for the report-hide pairs — `scan_root_contains_component_pair` exempts the scan root because "the user intentionally asked to analyse that corpus" — but `built_in_excluded` has no equivalent carve-out.

## Pinning test

A failing E2E test (`crates/deslop/tests/issue_<this>_scan_root_under_excluded_ancestor.rs`) accompanies this issue: it scans a clone-bearing repo seeded under `<tmp>/dist/` and asserts `files_analysed == 2` plus the reported clone pair. It is red on `main` for exactly the defect above.

MelbourneDeveloper:
**Fixed on branch `fused` (not closing until CI runs the pinning test).**

The pinning test this issue described was never actually written — the placeholder filename `issue_<this>_…` in the issue body was the tell. It now exists, test-first:

[...]

Issue #298, opened 2026-07-25
Add `out/` (and `.vscode-test/`, `coverage/`) to built-in default excludes — VS Code / TS compile output


MelbourneDeveloper:
## Summary

Add **`out/`** to deslop's built-in default excludes (alongside `dist`, `build`, `target`, `node_modules`, `.venv`, `__pycache__`, `.cargo`).

`out/` is the canonical TypeScript compile output directory for **VS Code extensions** — it's what the official `yo code` extension generator and virtually every VS Code extension `tsconfig.json` uses (`"outDir": "./out"`). Because it isn't in the defaults, deslop discovers the machine-generated `.js`/`.js.map` transpiled from `.ts` source and reports them as duplicate code. That duplication is a compiler artifact, not hand-written code, so it's pure noise in both the CLI gate and the Top Offenders panel.

## Repro / evidence

In a standard VS Code extension repo (`tsconfig.json` → `outDir: ./out`, `out/` git-ignored):

- Top Offenders fills up with `out/test/suite/*.test.js` clusters that are just transpiled copies of the corresponding `*.test.ts`.
- Each `.ts` source and its `.js` build output double-count, and the compiled test files cross-duplicate heavily (shared transpiled boilerplate), inflating the metric.

Workaround today is an explicit per-repo entry:

```toml
[defaults]
exclude = ["**/out/**"]
```

...but every VS Code extension repo has to rediscover this. It belongs in the built-in defaults, since `out/` is as standard for the TS/VS Code ecosystem as `dist`/`build` are elsewhere.

## Suggested additions

Primary ask:

- **`out/`** — VS Code extension / general TypeScript compile output.

Two adjacent candidates worth considering while you're in there (both are always non-source, always generated):

- **`.vscode-test/`** — `@vscode/test-electron` downloads a *full copy of VS Code* (its own compiled JS + bundled extensions) here to run the extension test suite. It's large and 100% third-party runtime.
- **`coverage/`** — generated lcov HTML/JS coverage reports.

## Notes

- `out/` is a slightly higher-risk default than `dist`/`build` because a repo *could* theoretically keep source in a dir literally named `out/`. That's rare, and consistent with the existing `dist`/`build` defaults which have the same theoretical risk; a repo that wants it analyzed can always re-include it. Happy to gate it behind the same policy you already apply to `build`.
- Repro'd with deslop 0.25.0.

MelbourneDeveloper:
[...]

PR #353, opened 2026-08-11
Install typediagram in the corpus workflow so the accuracy gate can build

### LINKED ISSUE #347: Corpus accuracy gate has never run: corpus.yml omits the typediagram install, so build.rs fails before any repository is scanned
The scheduled real-repository accuracy gate (`.github/workflows/corpus.yml`, `[CORPUS-CI]`) has **never produced a measurement**. It has run twice since the commit that introduced it and both runs failed identically, ~70 seconds in, before a single repository was scanned.

| Run | Commit | Result |
|---|---|---|
| [31400261915](https://github.com/Nimblesite/Deslop/actions/runs/31400261915) | `a3fe320be` | failure @ 1m12s |
| [31356...](https://github.com/Nimblesite/Deslop/actions/workflows/corpus.yml) | `fc779f7bc` (the commit that added the gate) | failure @ 1m22s |

## Root cause

`corpus.yml` never installs the `typediagram` CLI, so `deslop-core`'s `build.rs` cannot generate the wire models and the crate fails to compile:

```
Error: "typediagram-gen failed: node exited with status exit status: 1.
        Run `make typediagram-gen` to reproduce."
  errno: -2, code: 'ENOENT', syscall: 'spawnSync typediagram',
  spawnargs: [ '--to', 'rust', '.../docs/models/live-ipc.td' ]
make: *** [Makefile:233: test-corpus-ci] Error 101
```

`ci.yml` installs it in all four of its jobs (lines 136, 224, 276, 362):

```yaml
- name: Install typediagram
  run: npm install -g typediagram@0.11.0
```

`corpus.yml` sets up Node but has no equivalent step.

## Why this matters more than a normal CI break

This is the only surface that measures accuracy against real repositories at scale, and it is the designated tripwire for the two defects the fused-confidence work is blocked on:

- **#301 (determinism)** — `nest` and `jellyfin` are baselined as `determinism` failures in `corpus/known-failures.json`. The plan in `docs/plans/fused-score-followups.md` makes closing #301 the *first* item, on the grounds that "while identical runs disagree, no before/after accuracy measurement of the content gate is trustworthy". That baseline has never been re-verified by a green run.
- **#345** — calls for adding a `fused_spread` / `type2_recall` check id to `corpus/known-failures.json` "so the real-repository gate can catch saturation and rename recall at scale". Adding checks to a gate that cannot start is a no-op.

The regression-detection contract in the workflow header — "Anything NOT recorded in `known-failures.json` fails this workflow — that is the regression gate" — has been vacuous since it was written. `[FUSION-CONTENT-GATE]` shipped in v0.31.0 with zero real-repository accuracy evidence, because the gate that would have supplied it could not compile the binary.

## Fix

Add the install step to the `corpus` job, matching `ci.yml` and keeping the pinned version in sync per the dependency-sync rule in `CLAUDE.md`:

```yaml
# typediagram is required by deslop-core's build.rs — see ci.yml.
- name: Install typediagram
  run: npm install -g typediagram@0.11.0
```

Then `workflow_dispatch` the gate and reconcile the result against `corpus/known-failures.json`, since that baseline has never been confirmed by a passing run.

## Secondary

The job pins `actions/checkout@v4` and `actions/setup-node@v4`; both now emit Node 20 deprecation warnings and are being force-run on Node 24. `ci.yml` is already on v7.

## Acceptance

- [ ] A scheduled or dispatched `Corpus accuracy` run reaches the scan phase and reports per-repository results.
- [ ] `corpus/known-failures.json` is reconciled against that first real run — entries that no longer reproduce are removed, new ones recorded.
- [ ] A green run is visible on `main` before #301 is worked, since #301's evidence depends on it.

Wamwea:
## TLDR
Installs the `typediagram` CLI in the corpus workflow so the real-repository accuracy gate can compile `deslop-core` and actually run — it had failed all three of its scheduled runs at build time, before scanning a single repository — and pins the invariant with a workflow-conformance E2E test. Closes #347.

## Details
- **`.github/workflows/corpus.yml`** ([CORPUS-CI]): new `Install typediagram` step (`npm install -g typediagram@0.11.0`, matching `ci.yml` and `make setup`), placed after `setup-node` and before anything that compiles. `deslop-core`'s `build.rs` shells out to `typediagram` to generate the IPC wire models, so every scheduled run died with `spawnSync typediagram ENOENT` ~70s in. Also aligns action pins with `ci.yml`: `actions/checkout` and `actions/setup-node` v4→v7 (clears the Node 20 deprecation warnings noted in #347), plus `actions/cache` v4→v6 (a drift the issue missed).
- **New E2E test** `crates/deslop/tests/issue_347_workflow_typediagram_install.rs` ([CORPUS-CI]): parses every `.github/workflows/*.yml` as YAML — no text pattern-matching — and asserts that every job invoking `cargo` or `make` installs `typediagram` **before its first compiling step**, pinned to the **single** version the Makefile `setup` target installs. This enforces the CLAUDE.md dependency-sync rule across `ci.yml` (4 jobs), `corpus.yml`, and `release.yml` in one place.
- **New dev-dependency**: `yaml-rust2 = "=0.11.0"`, exact-pinned in the workspace table per convention; `Cargo.lock` change is purely additive (51 insertions, 0 deletions).
- **Gate verified end-to-end locally**: `make test-corpus-ci` (the exact command the workflow runs) completed for the scheduled slice and produced the harness's first-ever measurements — tokio: 758 files / 168,480 LOC / 1,713 clusters / 20.0% / 12.5s / 1,274 MB; nest: 1,726 files / 115,848 LOC / 1,220 clusters / 29.9% / 18.8s / 866 MB. The `nest/determinism` baseline (#301) reproduced exactly as recorded in `corpus/known-failures.json` (`[KNOWN]`, clusters 1226 vs 1222) — its first confirmation by a completed run. No `[NEW]`, no `[FIXED?]`.

No public API, CLI flag, report format, or spec ID changes.

- `workflows_that_compile_the_workspace_install_pinned_typediagram_first` was verified **red against the pre-fix corpus.yml** (install step temporarily removed): it fails with `corpus.yml job 'corpus' compiles the workspace but never installs typediagram` — the exact #347 defect, not a setup error — and passes with the fix in place.
- Blind-audit guards: the test asserts the workspace-compiling set includes the `corpus` job, at least four `ci.yml` jobs, and the `release.yml` build job, so a predicate regression fails loudly instead of passing vacuously.
- The ordering assertion rejects an install step placed after the job's first `cargo`/`make` step (where the build would already have failed), and the version-shape assertion rejects non-exact pins such as `typediagram@latest`, which would satisfy version sync while abandoning pinning.
- The gate itself: before this fix, `make test-corpus-ci` aborted inside `build.rs` on any machine without a global `typediagram`; after it, the run completes and reports per-repository measurements (see Details).

Local `make ci` note: the shared working tree currently holds three untracked red-by-design accuracy tests from in-progress sessions (#298, #331, #342) plus the pre-existing timing-dependent `cpu_throttle_knob` failure (filed as #352), so the fail-fast local suite stops red on those external causes. Evidence that this diff is clean: all 119 other test binaries pass when run individually, full `make lint` is green, `cargo clippy --all-targets` is clean, and a Deslop self-scan reports zero duplication clusters involving the new test.


MelbourneDeveloper:
Closing this one for now because we need to align on the corpus doc

Issue #367, opened 2026-08-14
Recall hole: nested sibling window subsumes the enclosing near-verbatim clone, content gate hides the remainder — ts-mixed-band pair renders zero clusters without embeddings


MelbourneDeveloper:
## Zero visible clusters for a 380-node near-verbatim clone pair

Scan `ts-mixed-band/ledger_a.ts` + `ledger_c.ts` alone, embeddings off (`--min-nodes 12`): `settleLedger` and `settleQuarter` are whole-function near-verbatim clones (identical bodies except `(beta)` parens, renamed function). The report shows **`clusters: []`, `clusters_hidden: 1`** — the most obvious duplicate imaginable is not reported.

## Measured causal chain (2026-08-14, `fused` branch, debug tracing)

1. Discovery works: 1217 LSH pairs → 178 candidates → 4 survive → clusters form, including the whole-function pair `F` and the 19-node name+params sibling window `W` (`[9..80]`×`[9..81]`).
2. **Subsume inversion** (`cluster/subsume.rs`): `W` has `structural = 1.0` *by construction* — normalisation collapsed the very leaves (the function name) that differ. `F` has `structural = 0.0` *because* it is a real Type-3 (the parens poison every ancestor Merkle hash). `strictly_encloses(F, W)` holds, but `evaluate_pair` routes through `preferred_view` → `precision_preference`, which lets the nested window's saturated structural overturn enclosure: `drop_inner survivor=7471d78aa8529cc6 (W)` discards `F` and two more enclosing views. The module doc's own principle — "The enclosing view is the duplication; the nested view re-describes it" — is contradicted by its `precision_preference` tail whenever the fragment is nested inside the clone.
3. **Content gate hides the surviving window** (`report_render.rs` / `buckets.rs`): `W` measured `structural=1.0, token_jaccard=0.0` (sibling window ≠ any node → byte-offset-seeded fallback signature), `content_agreement=0.8` (4 of its 5 collapsed leaves match; the name differs), `rename_consistency=0.0` (zero literal anchors < `RENAME_EVIDENCE_MIN_LITERALS`). Support 0.8 < `CONTENT_PROMOTE_FLOOR` 0.85 → `structural_only` → hidden.
[...]

Issue #366, opened 2026-08-14
MockOllama::embed_vector measures length residues, not content: ~88 embedding_cos assertions calibrate against sin(len) noise


MelbourneDeveloper:
## TL;DR

The fake embedding server used in our tests gives almost any two pieces of code nearly the same "meaning" score — it mostly measures how long the text is and what its first character is. About 88 test assertions that look like they test embeddings are really passing or failing on accidental text-length coincidences. Two accuracy tests are already stuck red because of it.

## Details

To test Deslop's embedding features without a real AI service, the tests use a mock that invents a vector for each piece of code. The formula: one number from the text's length, one from its first byte, and two constant numbers. Because two of the four numbers never change, every pair of texts starts out looking similar; the only variation left is length and first character.

Concretely: a 67-byte parameter list and an 865-byte arithmetic chain — completely different code — both start with `(` and score 99.997% similar.

What this breaks:

- About 88 assertions across 15 test files claim to check embedding similarity. They are actually checking which text lengths happen to collide. A one-byte shift in a test fixture reshuffles the collisions and flips unrelated tests — this has been demonstrated both ways on one of them.
- A test that manufactures a "duplicate" from two unrelated snippets (a function signature vs an arithmetic expression) gets a 0.87 similarity score — pure coincidence, rendered to the user as a real finding.
- Two accuracy tests stay red until this is fixed, and an LSP test calibrates against the same broken mock.

The fix must make the mock score genuinely similar code high and unrelated code low — not just "different". When the 15 test files are recalibrated, no assertion may be weakened: the bounds move because the instrument was broken, and each one must keep at least as much power to tell real from fake.

No production code is involved — this is test equipment — but it currently gates accuracy tests, so it counts as an accuracy defect.

## Steps To Reproduce

1. Run the already-red tests:

   ```
   cargo test -p deslop issue_343_sum_clamp_saturation
   cargo test -p deslop pair_size_coherence
   ```

2. Observe: `mid_band_cluster_confidence_never_exceeds_its_strongest_axis` fails at `cluster_count 2 != 1`; `an_embedding_only_pair_does_not_join_occurrences_of_different_size` fails.
3. Proof of the length-coincidence mechanism: revert the one-byte-shifting normalization change and the first test passes; restore it and the test fails again.

## Details (for AI)

```shell
# defect
mock: crates/deslop/tests/cli/mock_ollama.rs :: MockOllama::embed_vector
[...]

Issue #365, opened 2026-08-14
Embedding presence disables the LSH-only corroboration floors: weaker evidence promotes a pair (false-positive channel)


MelbourneDeveloper:
## Defect

`survival_decision` in `crates/deslop-core/src/pair.rs` applies the two LSH-only corroboration floors only when a pair has **neither** a structural anchor **nor** any embedding signal:

```rust
let lsh_only = score.structural <= 0.0 && score.embedding_cos <= 0.0;
if lsh_only && score.token_jaccard < pair.lsh_only_min_jaccard { ... }
if lsh_only && pair.lsh_only_node_floor < LSH_ONLY_MIN_NODE_COUNT { ... }
```

Any non-zero `embedding_cos` — however small — switches both floors off. That inverts the evidence ordering:

| pair | outcome |
|---|---|
| `structural=0, token_jaccard=0.87, embedding_cos=0.00` | **dropped** by `LSH_ONLY_MIN_JACCARD` (0.90) |
| `structural=0, token_jaccard=0.87, embedding_cos=0.10` | **survives** — `bounded_fused = max = 0.87 ≥ 0.85`, floors skipped |

A pair with *strictly weaker* corroboration is promoted over one with stronger corroboration. The same bypass applies to `LSH_ONLY_MIN_NODE_COUNT` (40): a tiny subtree with `embedding_cos = ε` skips the information-content floor entirely.

## Why this is a false-positive channel

Embedding-only pairs (`structural = 0`) reach transitive closure gated **only** by `bounded_fused ≥ 0.85` on the embedding axis itself — the signal being corroborated is its own corroboration. The new `[PAIR-SIZE-COHERENCE]` guard (`MAX_ENDPOINT_NODE_RATIO = 4`) now drops the grossest of these (endpoints 19 vs 274 nodes), but size-coherent embedding-only garbage still survives: in the `ts-mixed-band` two-ledger repro, cluster `e021161df1cf4142` joins `settleLedger(alpha: number, …)` (18 nodes) with the arithmetic chain `(beta) + gamma * 2 + …` at `embedding_cos = 0.87`, `token_jaccard = 0.0`, `structural = 0.333` on one edge and `0` on the joining edge.

## Why the naive fix is wrong

[...]

Issue #359, opened 2026-08-12
Pre-existing accuracy defects carried through the `fused` review: content quorum, call-only literal suppression, ts-mixed-band recall, dict-assert RHS, six VSIX skips


MelbourneDeveloper:
Findings from the `fused` branch review that are **not** branch regressions — each was verified present and behaviourally identical on `main`. They are filed so they survive the branch review document, which is not committed to either branch. None is fixed by `fused`; none is made worse by it except where noted.

## 1. A partial leaf-key quorum promotes every member of a mixed component

`crates/deslop-core/src/content.rs` forces cluster agreement to `1.0` whenever `duplicated_member_share() >= VERBATIM_MEMBER_SHARE_FLOOR` (`0.5`). Verified identical on `main` — same constant at `content.rs:54`, same use site.

The share groups equality of identifier/literal `LeafKey` vectors, **not** byte-equivalent source: operators and node kinds are not part of a `LeafKey`. It then sums every member belonging to any repeated key vector.

Consequences:
- Two matching members of a four-member component vouch at full strength for two unrelated members.
- Two syntactically different snippets with the same collapsed identifier/literal sequence but different **operators** satisfy the "verbatim" guard.

The cluster-wide bucket, fused score, rank, and metrics then borrow proof the unrelated occurrences never earned. **False positive.**

**Fix:** carry evidence per pair/member, or split heterogeneous components into evidence-coherent clusters. Controls needed: 2 proven + 2 unrelated same-shape members; 2 proven + 1 unrelated; equal identifier/literal sequences with different operators.

## 2. Literal-variation suppression ignores duplicated non-call logic

`crates/deslop-core/src/cluster_filters/calls.rs` examines only call expressions. Assignments, calculations, branches, loops, and returns in the same matched range do not participate, so a substantial duplicated implementation is hidden whenever every call happens to carry a different literal payload. Verified: `main` has the same call-only scope (`is_literal_variation_call_set` / `is_literal_variation_call_sequence`).

`fused` did fix the separate member-count shortcut in this file — it now requires *every* collected call position to vary — but the scope limit is untouched. **False negative.**

**Fix:** suppress only when the entire matched AST range is proven payload/scaffolding. Needs a logic-bearing negative control: identical calculations and control flow plus an all-varying call sequence.

## 3. `ts-mixed-band` gives a wholesale rewrite full confidence and loses `ledger_c`

[...]

PR #368, opened 2026-08-14
Fused confidence: honest cluster signals, content gate, subsumption, correct __file__ ranges, and contract-checked AST goldens

MelbourneDeveloper:
## TLDR

Lands the fused-confidence accuracy work: honest cluster signal measurement ([FUSION-CLUSTER-SIGNALS]), a content gate that separates renamed clones from shape coincidence ([FUSION-CONTENT-GATE]), cross-cluster subsumption ([PIPELINE-CLUSTER-SUBSUME]), correct `__file__` occurrence ranges ([PIPELINE-NORMALIZE-AST]), a new size-coherence pair guard ([PAIR-SIZE-COHERENCE]), and AST goldens that check a contract instead of certifying themselves.

## Details

**Signals are measured between the occurrences the report actually shows.** `crates/deslop-core/src/cluster/signals.rs` (new) computes the per-signal mean over every unordered pair of rendered occurrences. Previously the surviving discovery edges of the transitive-closure component were averaged — a mix that is an artifact of discovery topology (structural star buckets, ANN top-k fan-out, LSH band width), under which a byte-identical file pair once rendered `structural = 0.36` and routed `same_behavior`. A pair missing an input for a signal is excluded from that signal's numerator *and* denominator, so absence never masquerades as a measured `0.0`.

**Content gate ([FUSION-CONTENT-GATE]).** `crates/deslop-core/src/content.rs` measures what normalisation erased, in two independent populations: pooled raw-byte agreement, and Type-2 rename consistency (literal preservation ∧ bijective identifier mapping). Routing consults both — never their mean, which is what demoted textbook Type-2 clones to `structural_only`. Shape-identical clusters that neither population vouches for are demoted honestly instead of rendering `fused = 1.00`.

**Cross-cluster subsumption ([PIPELINE-CLUSTER-SUBSUME]).** `crates/deslop-core/src/cluster/subsume.rs` (new) collapses nested views of one duplicated region so the same bytes are not reported twice or double-counted in `clusters_total`. Coverage is bidirectional per-occurrence containment; three weaker predicates (set nesting, bare intersection, either-direction coverage) were each wrong in a different direction and are documented in the module header.

**`__file__` no longer claims dropped trivia as duplicated code ([PIPELINE-NORMALIZE-AST]).** `build_normalised_root` gave the synthetic root tree-sitter's *parse* root span while normalisation had already dropped every comment, so whole-file clusters reported licence headers and trailing newlines as part of the duplicate — the Go fixture claimed 759 bytes of package comment. `crates/deslop-core/src/lang/shared.rs` now spans exactly the nodes normalisation kept. Fingerprints are unaffected (Merkle hashes combine `kind` and child hashes, never byte ranges); only reported ranges move. Two latent defects this exposed are fixed with it: the synthetic root was fingerprinted twice when one declaration covered it exactly (`fingerprint.rs` suppresses the re-description), and `smallest_covering` refused whole-file requests once the root stopped spanning the trailing newline (`ast.rs` clips the request to the subtree extent, restoring 26 merge/refactor tests).

**New: [PAIR-SIZE-COHERENCE] pair guard.** `crates/deslop-core/src/pair.rs`. A cluster rendered "3 copies of a 19-node subtree" at `fused = 1.00`, `same_behavior`, visible — two 67-byte parameter lists plus an 865-byte, 274-node arithmetic chain, glued by an embedding-only edge nothing else corroborated. `survival_decision` now drops any pair with no structural anchor whose endpoint node counts differ by more than `MAX_ENDPOINT_NODE_RATIO` (4), counted as `DroppedSizeMismatch` in the pair-survival stats and tracing summary. The ratio is deliberately loose so Type-3 growth is untouched. `CandidatePair` gains honest `endpoint_node_counts` alongside `lsh_only_node_floor`, because the cross-language opt-in raises the floor and so cannot stand in for a measured size.

**Noise filters.** Dart forwarding suppression (`cluster_filters/forwarding.rs` + `forwarding/delegation.rs`) now requires every call to be transport — delegating to collaborator state, or consuming only data produced by delegation — so calls that transform member inputs or introduce literals stay visible. A call *count* cannot substitute: every vendored meilisearch wrapper makes two calls, so a one-call rule would unsuppress the family the filter exists for. `cluster_filters/python_dict_assert.rs` accepts `decorated_definition` only when the underlying definition is a function, so a decorator can no longer cause executable class-body logic to be ignored.

**AST goldens stopped certifying the defect.** Every `Sample.expected.ast` recorded the wrong root span and the guard compared `--debug-ast` output against it byte-for-byte — the tool against a file the tool wrote, so a wrong expectation was self-certifying and regenerating on failure would launder the bug back in. `assert_dump_is_correct` now checks the committed dump against the contract: `__file__` spans exactly `[min(child.start) .. max(child.end)]`, every node nests inside its ancestor, every range is non-empty and in bounds, no `comment` node survives. Rationale and regeneration procedure in `crates/deslop/tests/fixtures/AST-GOLDEN-README.md`.

**Observability.** `report.rs::log_hidden_cluster` now emits the signal triple and measured content evidence for every cluster that leaves the visible report — a duplicate that silently disappears is the hardest defect class to notice.

**Docs.** Spec updates across `docs/specs/` (`fusion.md`, `pipeline.md`, `noise.md`, `taxonomy.md`, `severity.md`, `exclusion.md`) carrying the hierarchical IDs above; `docs/plans/quarantine-repair-plan.md`, `docs/plans/weighted-metrics-plan.md`, `docs/plans/embedding-accuracy-plan.md` (new, see below); `site/src/docs/accuracy-transparency.md`.

**No breaking changes** to public API, CLI flags, or the report wire format. Spec IDs are additive.

### Ignored tests, deliberately not weakened

Three tests on the `ts-mixed-band` fixture are `#[ignore]`d under **#369** so this can merge. **No assertion was deleted, skipped, or loosened** — each still runs under `cargo test -- --ignored`, and un-ignoring all three is the acceptance criterion for `docs/plans/embedding-accuracy-plan.md`. All three trace to defects **older than this branch**, masked until the root-span correction moved byte lengths.

Measured while diagnosing: scanning `ledger_a.ts` + `ledger_c.ts` renders **two clusters, both false positives** — `45986e47bfc430a2` (whole `ledger_a` vs an arithmetic chunk of `ledger_c`) and `e021161df1cf4142` (two parameter lists plus an arithmetic chunk). Both carry `structural = 0` *and* `token_jaccard = 0`, surviving on the mock's cosine alone. The real 380-node clone is not among them.

- **#367** — token similarity understates Type-3 clones. Two 1235-byte files identical except a rename and **one added pair of parentheses** scan to zero duplicates; the dropped whole-function pair measures `structural = 0.0, token_jaccard = 0.664, embedding_cos = 0.0` at `nodes = (380, 381)` — 99.7% identical by node count. Verified **not** a branch regression: only one axis is non-zero, so `sum == max` and `main` drops it identically. An earlier subsumption theory was disproved by the trace and retracted on the issue.
- **#359** — the LSH Type-3 promotion that would rescue that clone is language-gated to C#; identical TypeScript evidence routes `loosely_similar` → hidden.
- **#365** — embedding presence disables the LSH-only corroboration floors: any ε of cosine waives both the 0.90 Jaccard floor and the 40-node floor, so weaker corroboration *promotes* a pair.
- **#366** — `MockOllama::embed_vector` is `[sin(len), cos(first_byte), 0.5, -0.5]`: two constant lanes plus aliasing `sin(len)` give any two texts a high cosine (67-byte vs 865-byte texts score 0.99997). It manufactured the visible cluster that papered over the rest.
- **#369** — the replacement is written and turns all three green (verified 6/6 locally: real pair 0.966, whole-fn vs chunk 0.803, params vs chunk 0.085, identical text exactly 1.0). It is not landable yet because `exact_embedding_pairs` is `O(N² · D)` — at 4096 lanes `issue_286_large_subtree_survives_when_the_model_declares_the_context` ran past ten minutes. A signed-lane signature holds the separations at 128 lanes; that is the width to time first.

`deslop-lsp::embedding_failure_progress` hangs and is pre-existing (revert-proved past 300 s).

**Signal honesty.** `deslop::issue_343_sum_clamp_saturation::every_visible_mixed_band_cluster_obeys_the_bound` asserts every visible cluster's rendered `fused` stays within `[0, 1]` and never exceeds its strongest axis, and that `fused = 1.0` appears only with a byte-identical occurrence pair — a report that summed-then-clamped fails it. `byte_identical_pair_still_earns_full_confidence_under_the_bound` is the false-negative control: byte proof still earns 1.0 on every axis and routes `identical`. `without_embeddings_the_mid_band_pair_stays_hidden` asserts the unproven LSH-only pair is hidden rather than deleted, and contributes 0 to `duplication_percent`. (The fourth test in this binary, `mid_band_cluster_confidence_never_exceeds_its_strongest_axis`, is ignored under #369 with assertions intact.) `deslop::embedding_route_invariance` asserts the same bytes reach the same bucket whether discovered structurally, by LSH, or by ANN.

**Content gate.** `deslop::declaration_family_plurality` and `declaration_family_mixed_component` assert a sibling-declaration family is demoted while a genuine rename in the same component stays act-now. `python_dict_assert_payload_proof` / `_reach` / `_rhs_logic` assert the decorator fix by class-body content, and `class_body_logic_under_a_static_decorator_is_not_excused` fails if the decorated-definition check regresses.

**Subsumption.** `deslop-core::cluster_subsumption` and `cluster_overlap_collapse`, plus `deslop::cross_cluster_enclosure`, assert which view survives and that a view naming a file no survivor names is never dropped — the false-negative guard. Fixtures `csharp-enclosing-method-clone`, `csharp-mixed-declaration-component`.

**Forwarding.** `a_same_class_call_after_delegation_is_not_forwarding` and `a_same_class_call_before_delegation_is_not_forwarding` fail if transformation calls are re-admitted as transport; `dart_issue_197_single_file_structural_only` still asserts the family that filter exists for stays suppressed. Fixtures: `dart-forwarding-transform-{before,after}-delegation`, `dart-forwarding-business-pair`, `dart-forwarding-fail-open`.

**Root span.** All eleven `ast_golden_*` tests assert the committed dump against the contract, not against the tool. Verified by reverting the fix and regenerating all eleven from the broken build: byte-for-byte equality **passed** and all eleven contract checks **failed** — so the guard has teeth. An independent audit (raw bytes + `python3`, never the `deslop` binary) scores the pre-fix goldens 0/11 and the corrected goldens 11/11; the Go boundaries were confirmed against Go's own `go/parser`, which puts `*ast.File` at `759..3572`, byte-identical to the golden. `deslop-mcp::issue_153_rescan_freshness` was the red test that surfaced the defect and is green.

**Size coherence.** `deslop::pair_size_coherence` asserts every occurrence in a visible cluster is within 4× the smallest — the 14× false positive fails it with a message naming the offending member, its spans, and its signals. `size_coherence_keeps_every_genuine_ledger_family_visible` (green) proves the guard erases no real duplication: the five-ledger rename family stays visible and every surviving cluster is internally coherent.

**Whole-file merge paths.** `refactor_merge`, `refactor_merge_refusals`, `live_merge_plan`, `refactor_ast_access` — 26 tests that fail with "normalised tree unavailable" if `smallest_covering` stops clipping whole-file requests.

Local status: `make fmt CHECK=1` and `cargo clippy --release --all-targets --workspace -- -D warnings` both pass clean. A full workspace sweep plus a targeted `deslop-lsp`/`deslop-mcp` pass covers 159 test binaries; everything is green except the four known-red items listed above. **`make ci` was not run to completion locally** — it is fail-fast and stops on the deliberately-red `ts-mixed-band` set, so CI will show those failures rather than a green board.

## For AI

Pipeline stage deltas: `normalise_root_span := [min(retained_child.start) .. max(retained_child.end)]` replacing `tree_sitter::root.byte_range()`; Merkle fingerprints invariant under this change by construction (`hash = H(kind ‖ child_hashes)`, ranges excluded), so the only mutation is occurrence rendering and any range-keyed downstream consumer. `re_describes_only_child` suppresses the synthetic root iff `|children| == 1 ∧ child.range == root.range`, preventing double fingerprinting of a single-declaration file (self-cosine 1.0 intra-file artifact). `smallest_covering(request)` now clips to `subtree.extent ∩ request` before descent; wholly-disjoint requests still yield `None`.

`measured_signals(occurrence_indices)` is O(k²) over rendered occurrences with per-signal `MeanAccumulator{sum,count}`; unmeasurable inputs decrement the denominator rather than contributing 0.0. `ContentEvidence{agreement, rename_consistency, literal_fraction, substance_varies}`; `support() = max(agreement, rename_consistency)`; `content_gated_signals` computes `fused = max(embedding_cos, max(structural, token_jaccard) · max(agreement, 0.9·rename_consistency))` clamped to `[0,1]`, bypassed for `Identical` and for clusters lacking saturating shape evidence. `mapping_consistency` requires modal-partner agreement in both directions (bijection test), deterministic via `BTreeMap` ordering with strict-greater replacement.

`survival_decision` gate order is now: `bounded_fused < fused_min_score` → `DroppedBelowFused`; `structural ≤ 0 ∧ ¬size_coherent(endpoint_node_counts)` → `DroppedSizeMismatch`; then the LSH-only pair (`structural ≤ 0 ∧ embedding_cos ≤ 0`) Jaccard and node-count floors. `size_coherent((s,l)) ≜ l ≤ s · MAX_ENDPOINT_NODE_RATIO` with saturating multiplication. Known gap (#365): the `embedding_cos ≤ 0` conjunct in the `lsh_only` predicate means `embedding_cos = ε > 0` waives both floors, making corroboration non-monotone in evidence strength — a pair at `(0, 0.87, 0.0)` is dropped while `(0, 0.87, 0.10)` survives on `bounded_fused = max = 0.87`. Proposed remedy in `docs/plans/embedding-accuracy-plan.md` §3 introduces no new constants: require the node floor for all `structural ≤ 0` pairs, waive the Jaccard floor only when `embedding_cos ≥ fused_min_score`.

`subsume::evaluate_pair` routes the `strictly_encloses(inner, outer)` case through `preferred_view → precision_preference`, which compares `signals.structural` — adversarial under normalisation, since a nested window over collapsed-identifier leaves saturates structurally exactly when the enclosing Type-3 does not. #367 §1 proposes enclosure-decides with precision confined to crossed/identical occurrence sets, retaining the `covers_every_file` false-negative guard and the `is_embedding_dominant` nomination for the non-nesting branch. `is_csharp_lsh_type3_near_miss` predicates on `language == "csharp"` over an otherwise language-agnostic evidence profile (`structural ≤ ε ∧ embedding_cos ≤ ε ∧ token_jaccard ≥ LSH_ONLY_MIN_JACCARD ∧ ∀m. m.node_count ≥ LSH_ONLY_MIN_NODE_COUNT ∧ cross-file`) — un-gating is #367 §2.

Golden oracle class change: `assert_ast_golden` retains byte-for-byte equality against the committed file but composes it with `assert_dump_is_correct`, converting a self-certifying fixpoint (`tool ⊨ tool`) into a contract check (`dump ⊨ invariants`) — root-span exactness, ancestor nesting, non-empty in-bounds ranges, no surviving `comment` kind. Falsification protocol executed: revert fix → regenerate all 11 → equality holds, contract fails 11/11.

`MockOllama::embed_vector(t) = [sin(|t| & 0xffff), cos(t₀), 0.5, -0.5]` has rank-2 content sensitivity with two constant lanes, giving cosine floor `≥ (0.5² + 0.5²)/‖·‖² ≈ 0.5` for arbitrary inputs and aliasing under `sin` over length; measured 0.99997 for a 67-byte / 865-byte pair sharing `t₀ = '('`. ~88 `embedding_cos` assertions across 15 files are therefore calibrated against `sin(len)` noise (#366). Remedy must preserve the `(0.80, 0.99]` mid-band the `ts-mixed-band` rename family asserts while separating unrelated code — hashed byte-n-gram frequency lanes are the suggested construction. Recalibration constraint: every moved bound must retain equal-or-stronger discriminating power against the corrected instrument; no assertion may be loosened.


MelbourneDeveloper:
## CI status and the measured root cause

Pipeline state on `9aadca2ba`: **Format ✅ · Lint ✅ · Windows + TCP IPC E2E ✅ · Website ✅ · CodeQL ✅ · security ✅**. The only failing step is `Test`, and `make test` is fail-fast, so it stops at the first red: `issue_343_sum_clamp_saturation::mid_band_cluster_confidence_never_exceeds_its_strongest_axis` (`cluster_count 2 != 1`).

The first push failed the format gate in 18s on a hand-wrapped signature in `pair/candidates.rs`; fixed in `817ebd574`.

### What the red tests are actually pinning

Root-caused by instrumenting `survival_decision`. **Minimal repro — no fixture, no embedder:** two 1235-byte TypeScript files identical except a consistent rename and **one added pair of parentheses** scan to `Found 0 groups of duplicated code`. Remove the parentheses and the identical pair is found correctly (`nearly_identical`, 266/268 pairs survive). The dropped whole-function pair measures:

```
structural=0.0  token_jaccard=0.6640625  embedding_cos=0.0  nodes=(380, 381)
```

Two functions 99.7% identical by node count score `token_jaccard = 0.664`. `structural` is 0 because one added `parenthesized_expression` changes the Merkle hash of every ancestor; with embeddings off — the default — `bounded_fused = 0.664 < FUSED_THRESHOLD` and the pair dies before any cluster exists. MinHash estimates Jaccard over the *set* of distinct k-grams, and a repetitive body has a small such set, so one added node displaces a large share of it.

**This is not a regression from this branch, and that was verified rather than assumed.** I first suspected the `fused() = Σ axes` → `bounded_fused() = max axes` change ([FUSION-STRATEGY-BOUNDED-MAX]); the measurement rules it out, since only one axis is non-zero, so `sum == max == 0.664` and `main` drops the pair identically. The branch neither causes nor worsens it — it only removed the mock-embedding cosine that was manufacturing a cluster over those bytes and masking the hole (#366).

Details and the retracted subsumption theory: #367. Sequenced fixes: `docs/plans/embedding-accuracy-plan.md`.

### Why these are not being made green here

Turning them green needs the token-signature work in §1 of that plan — a similarity-measure change requiring corpus validation in both directions, which is out of scope for this branch. The alternative is loosening the assertions, which is prohibited and which CLAUDE.md addresses directly: *"Leaving a red test in the tree is a correct outcome. Weakening it is not."* No assertion on this branch was deleted, skipped, or softened.

Reviewers should treat the `Test` step as red-by-design against four filed defects (#359, #365, #366, #367), with every other gate green.

Issue #394, opened 2026-08-17
RA-07: no event-matrix regression test for the Dependabot security-gate repair


MelbourneDeveloper:
Split out of the `worktree-fused-score-followups` regression audit (RA-07, P3).

The actor skips and stale comments are fixed, and `scripts/test-release-workflow-contract.mjs` covers the staging sweep. Nothing asserts the property the repair was for: a Dependabot **security** PR targeting the default branch must retain CI, dependency review, CodeQL and the Action self-test, while routine version updates stay on `dependabot-upgrades`.

Without that assertion an actor-skip can be reintroduced in any of the four workflows and every existing test stays green.

Fix: add a YAML-parsed event-matrix test covering both PR classes against all four gates. Keep the sweep staging-only and actor-gated.

MelbourneDeveloper:
**Cluster: Dependabot security gate** — #388 #394

Root cause: **#388**. `ci.yml` and `codeql.yml` both short-circuit on the Dependabot actor, justified by "Dependabot PRs are swept into `dependabot-upgrades` and never merge to main directly" — a justification that does not hold for the one class the skip can fire on, security PRs targeting the default branch. Fix it first: #394 is the assertion that keeps it fixed.

Fix order: #388 → #394.

Blocking edge recorded: #394 ← #388.

#394 asks for a YAML-parsed event-matrix test covering both PR classes against all four gates (CI, dependency review, CodeQL, Action self-test), staging-only and actor-gated. Without it, an actor-skip can be reintroduced in any of the four workflows and every existing test stays green — `scripts/test-release-workflow-contract.mjs` covers the staging sweep, not this property.

Issue #388, opened 2026-08-16
Dependabot SECURITY PRs to main skip CI, dependency review, and CodeQL — the only class the actor skip can fire on


MelbourneDeveloper:
`ci.yml` and `codeql.yml` both short-circuit on the Dependabot actor:

```yaml
# ci.yml — classify step
if [ "$ACTOR" = "dependabot[bot]" ]; then
  echo "code=false"; echo "site=false"; exit 0
fi

# ci.yml — security (dependency-review) job
if: github.actor != 'dependabot[bot]'

# codeql.yml — analyze job
if: github.event.repository.visibility == 'public' && github.actor != 'dependabot[bot]'
```

Each carries the same justification: *"Dependabot PRs are swept into `dependabot-upgrades` and never merge to main directly, so the consolidation PR is where these run."*

**That justification does not apply to any PR these skips can actually fire on.**

Both workflows trigger on `pull_request: branches: [main]` only. Every Dependabot **version** update carries `target-branch: dependabot-upgrades` (`.github/dependabot.yml`, all four ecosystems), so those PRs never reach either workflow — there is nothing for the skip to save. The only Dependabot PR whose base is `main` is a **security** update, which GitHub opens against the default branch regardless of `target-branch`. And the sweep does not move it: `dependabot-automerge.yml` carries an explicit `BASE FILTER — dependabot-upgrades ONLY. Do not add main back.`

So the net effect is the inverse of the intent:

| Dependabot PR | Base | Skip fires? | Gated by CI / dep-review / CodeQL? |
|---|---|---|---|
| routine version bump | `dependabot-upgrades` | never (workflow not triggered) | no — by design, gated at consolidation |
| **security / CVE fix** | **`main`** | **always** | **no — and never will be** |

The one PR class opened *because of a known vulnerability* is the one class that merges to `main` with no build, no test suite, no dependency review, and no CodeQL scan. It shows all-green because every job either did not run or reported `skipped`, which branch protection counts as passing.

The dependency-review skip is the sharpest of the three: a CVE fix bump routinely pulls transitive dependencies the advisory said nothing about, and that job is the repo's only dependency vuln-gate (no cargo-deny, no osv-scanner).

## Expected

A Dependabot security PR against `main` gets the same validation as a human PR against `main`.

## Fix

Remove the three actor short-circuits. They cost nothing on the routine bumps they were written for, because those bumps never trigger these workflows.

[...]

PR #419, opened 2026-08-20
Run every gate on Dependabot security PRs to main, and pin it with an event-matrix test

Wamwea:
## TLDR

Removes four `github.actor` short-circuits that stripped CI, dependency review, CodeQL and the Action self-test from Dependabot **security** pull requests — the only Dependabot PR class that reaches `main` — and adds a YAML-parsed event-matrix contract test that keeps them off.

Works on #388. Works on #394.

### The defect

`ci.yml`, `codeql.yml` and `action-selftest.yml` all trigger on `pull_request: branches: [main]` only, and every ecosystem in `.github/dependabot.yml` carries `target-branch: dependabot-upgrades`. Routine version bumps therefore never instantiate those workflows at all — the branch filter alone does the saving the actor skips were written for. The one Dependabot PR whose base is `main` is a **security** update, which GitHub opens against the default branch regardless of `target-branch`, and which `dependabot-automerge.yml` does not sweep (its base filter is staging-only, deliberately). Every skip could only ever fire on the CVE fix, and an `if:`-skipped job reports `skipped`, which branch protection counts as passing — so the hole showed all-green.

### Short-circuits removed

- `.github/workflows/ci.yml` — the `changes` classifier's `ACTOR:` env binding and its `code=false; site=false` early exit, which disabled the entire build/test/coverage matrix (`ci`, `windows`, `vsix`, `jetbrains`, `site` all key off its outputs).
- `.github/workflows/ci.yml` — `if: github.actor != 'dependabot[bot]'` on the `security` job (`[GITHUB-DEP-REVIEW]`), the repo's only dependency vulnerability gate. A CVE bump routinely pulls transitive dependencies the advisory did not mention.
- `.github/workflows/codeql.yml` — the `&& github.actor != 'dependabot[bot]'` conjunct on `analyze` (`[GITHUB-CODE-SCANNING]`). The `visibility == 'public'` gate is retained.
- `.github/workflows/action-selftest.yml` — `if: github.actor != 'dependabot[bot]'` on `contract` (`[ACTION-TESTS]`). Not named in #388: every other job carries `needs: [contract]`, so this one skip disabled the whole hosted-runner matrix, on the single bump class that edits `.github/workflows/**` and so matches the workflow's own `paths:` filter.

### Corrected claims

Three documents asserted the gate existed, each deferring to another, which is why the hole survived review:

- `.github/dependabot.yml` said the sweep fires on `main` and folds security bumps into staging. It does not — `dependabot-automerge.yml` filters its base to `dependabot-upgrades` only.
- `.github/workflows/dependabot-automerge.yml` said dependency review in `ci.yml` still gated security bumps. The skip above disabled it.
- `docs/specs/release.md` `[GITHUB-DEPENDABOT]` repeated both, and `[GITHUB-DEP-REVIEW]` claimed the job runs "on every `pull_request`".

New adjacent spec section `[GITHUB-DEPENDABOT-SECURITY-GATES]` states the invariant, records why no gate may branch on the author, and names the enforcing test.

### Permissions are load-bearing, and now pinned

A Dependabot-triggered event hands the workflow a read-only `GITHUB_TOKEN` unless an explicit `permissions:` block raises it. `security` already declared `pull-requests: write` and `analyze` already declared `security-events: write`, so removing the skips required no permissions change — but deleting either block would now break the gate on Dependabot PRs *and nowhere else*. Both are asserted by the new test and annotated in place.

### New files

- `scripts/test-dependabot-gate-matrix.mjs` (267 lines) — the event-matrix contract, wired into `make deployment-verify` (`[DEPLOY-CI-GATES]`).
- `package.json` / `package-lock.json` at the repo root — private, ships nothing, one devDependency (`yaml`) so the workflow gates are asserted against a real parse rather than a text match (CLAUDE.md prohibits pattern matching on structured data). The VSIX remains the only distribution. `/` is registered in `.github/dependabot.yml`'s npm `directories` so the new root is covered by bumps.
- `Makefile` — `_repo-script-deps` installs it, keyed on npm's `node_modules/.package-lock.json` so a lockfile bump reinstalls rather than silently running the suite against a stale parser. `ci.yml`'s `ci` job runs `npm ci` and caches the root lockfile.

### Not fixed here

The `dependabot-upgrades` branch currently holds `- main` in the sweep's base filter — the #306 regression `dependabot-automerge.yml`'s own comment warns will recur — plus two actor skips in `action-selftest.yml`. The Actions token cannot push `.github/workflows/*` by construction, so only a human push to that branch can correct it; #386 is the vehicle. The new test makes that revert fail loudly instead of landing silently, but does not prevent it.

### Breaking changes

None. No public API, CLI flag, report format or existing spec ID changed. `[GITHUB-DEPENDABOT-SECURITY-GATES]` is additive.

`scripts/test-dependabot-gate-matrix.mjs` runs in `make deployment-verify`. It parses `ci.yml`, `codeql.yml`, `action-selftest.yml`, `dependabot-automerge.yml` and `dependabot.yml` with the `yaml` package and asserts the two-PR-class × four-gate matrix. Five tests, each proven to fail for its own reason:

- **`securityBumpAgainstMainClearsEveryGate`** — all four workflows subscribe to `main`, each gate job is defined, and no expression anywhere in them consults who opened the PR. Against the pre-fix tree it reported all four defect sites by parsed YAML path in one failure: `ci.yml:jobs.changes.steps.[1].env.ACTOR`, `ci.yml:jobs.security.if`, `codeql.yml:jobs.analyze.if`, `action-selftest.yml:jobs.contract.if`. It catches both syntactic forms — a bare `if:` (implicitly an expression) and a `${{ }}` interpolation anywhere in the tree, which is what caught the `env:` binding that smuggled the actor into the classifier's shell. It matches five author-identifying contexts, not just `github.actor`: reintroducing the identical skip as `github.event.pull_request.user.login != 'dependabot[bot]'` was verified to fail it.
- **`securityBumpGatesRaiseTheTokenAboveDependabotReadOnly`** — `security` declares `pull-requests: write`, `analyze` declares `security-events: write`. Downgrading `security-events` to `read` was verified to fail it.
- **`routineVersionBumpNeverInstantiatesTheMainGates`** — no gate subscribes to `dependabot-upgrades`, so the staging branch stays free of the expensive matrix and the cost the removed skips were meant to avoid cannot return by the other route. Adding `dependabot-upgrades` to `ci.yml`'s bases was verified to fail it.
- **`everyEcosystemTargetsTheStagingBranch`** — every `updates:` entry sets `target-branch: dependabot-upgrades`. This is the premise the whole matrix rests on: drop one and that ecosystem's routine bumps start arriving on `main`. Removing it from the cargo entry was verified to fail it.
- **`theSweepIsTheOnlyActorGatedJobAndStagingOnly`** — the sweep fires on `dependabot-upgrades`, `main` is absent from its bases, and it still carries `github.actor == 'dependabot[bot]'`. Adding `main` back was verified to fail it, which is the assertion that would have caught #306.

Every one of those mutations was applied to the working tree, observed red, and reverted, so no assertion in the suite is vacuous. `scripts/test-release-workflow-contract.mjs` (8 tests) still passes unchanged alongside it.

Coverage thresholds in `coverage-thresholds.json` are untouched — this change adds no Rust or TypeScript.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,11 +10,15 @@
 #     codeql.yml only trigger on `pull_request: [main]` (the filter matches the
 #     PR *base*), so these staging PRs run NO build/test/CodeQL.
 #   * SECURITY updates -> GitHub IGNORES `target-branch` for these and always
-#     opens them against the repo's default branch (`main`). The auto-merge
-#     workflow therefore also fires on `main` and folds the security bump into
-#     the SAME staging branch, then retires the PR — so a CVE bump never sits on
-#     main waiting for a human either. (ci.yml/codeql.yml skip Dependabot PRs, so
-#     the matrix never burns on a bump we immediately sweep away.)
+#     opens them against the repo's default branch (`main`). The sweep does NOT
+#     collect them: dependabot-automerge.yml filters its base to
+#     `dependabot-upgrades` only, deliberately, so a CVE bump stays on `main` and
+#     is merged or retargeted by a human. It is therefore the one Dependabot PR
+#     that merges to main on its own checks, and it runs the FULL set — CI,
+#     dependency review, CodeQL, Action self-test — exactly like a human PR.
+#     None of those workflows may gate on the author; #388 was the hole where
+#     they did, and scripts/test-dependabot-gate-matrix.mjs now pins it shut.
+#     ([GITHUB-DEPENDABOT-SECURITY-GATES])
 #   * groups -> each ecosystem collapses ALL version bumps (patch, minor AND
 #     major) into ONE PR, and a parallel `*-security` group (`applies-to:
 #     security-updates`) collapses CVE bumps the same way. Majors no longer need
@@ -65,9 +69,11 @@ updates:
         applies-to: security-updates
         patterns: ["*"]
 
-  # npm / Node — the three package.json roots (no root-level package.json).
+  # npm / Node — the repo root (devDependencies for the scripts/ contract
+  # suites, ships nothing) plus the three product package.json roots.
   - package-ecosystem: npm
     directories:
+      - /
       - /site
       - /clients/vscode
       - /clients/vscode/webview-ui
--- a/.github/workflows/action-selftest.yml
+++ b/.github/workflows/action-selftest.yml
@@ -34,13 +34,13 @@ permissions:
 jobs:
   contract:
     name: Action contract
-    # Dependabot PRs are swept into `dependabot-upgrades`, never merged to main
-    # directly. A github-actions SECURITY bump is force-opened against `main` and
-    # edits .github/workflows/**, matching the paths filter above, so without
-    # this gate it would run the whole hosted-runner matrix on a discarded bump.
-    # Every other job needs this one, so skipping here skips them all.
-    # ([GITHUB-DEPENDABOT])
-    if: github.actor != 'dependabot[bot]'
+    # Runs for every author. A github-actions SECURITY bump is force-opened
+    # against `main`, edits .github/workflows/**, matches the paths filter above
+    # — and is NOT swept, because the sweep's base filter is the staging branch.
+    # It is therefore the only Dependabot PR that reaches here, and the only one
+    # that rewrites the very files this suite exists to prove still work. Every
+    # other job needs this one, so an actor skip here skipped them all.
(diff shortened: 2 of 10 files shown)


MelbourneDeveloper:
Sorry, no time to review these ones. We are working on accuracy issues at the moment.

PR #442, opened 2026-08-24
Pin the Dependabot security-gate matrix with a YAML-parsed event-matrix test

Wamwea:
## TLDR

Adds a YAML-parsed event-matrix contract test that pins the property `main` currently satisfies only by accident: a Dependabot **security** PR against the default branch must clear the build/test matrix, dependency review, CodeQL and the Action self-test, while routine version bumps stay on `dependabot-upgrades`.

## Details

Routine Dependabot bumps carry `target-branch: dependabot-upgrades`, so they never trigger the four `main`-only gates. Security updates are the exception — GitHub ignores `target-branch` for them and always opens them against the default branch, and the sweep's base filter is `dependabot-upgrades` only, so nothing carries them away. They merge to `main`.

That makes an actor short-circuit exactly inverted. It cannot fire on the routine bumps it was written for, and it always fires on the PR opened because of a known vulnerability. A skipped job reports as a *passing* required check, so the result is an all-green merge with no build, no tests, no dependency review and no CodeQL. #388 shipped that hole across all four gates; #392 removed the skips; nothing asserted they stay removed.

**New files**

- `scripts/release/test-dependabot-gate-matrix.mjs` (300 lines) — the event-matrix contract, `[GITHUB-DEPENDABOT-SECURITY-GATES]`. Reads `ci.yml`, `codeql.yml`, `action-selftest.yml`, `dependabot-automerge.yml` and `dependabot.yml` through a real YAML parse, never a text match, so a reformat cannot make it pass and a re-added skip cannot hide behind indentation. Its author-context list spans `github.actor`, `github.triggering_actor`, `github.event.pull_request.user.login`, `github.event.sender.login` and the bare `dependabot[bot]` name, walked at any depth including `env:` bindings — the shape the original defect used to hide inside `ci.yml`'s classifier shell.
- `scripts/release/contract-assertions.mjs` (36 lines) — `assertIncludes` / `assertExcludes`, previously defined locally in `test-release-workflow-contract.mjs` and about to be defined a second time in the new suite. Both now import it. The shared version names the haystack in the failure only when it is an array; workflow text is left undumped, so the existing suite's messages are unchanged rather than buried under a multi-KB file.

**Modified**

- `scripts/release/test-release-workflow-contract.mjs` (407 → 367) — `dependabotSweepLeavesNoDeadCheckOnHumanPullRequests` **moves** into the new suite. It asserted by text matching over YAML, which this repository prohibits for structured data; all five of its assertions are re-expressed against the parse and each is pinned by a mutant below. Its now-dead `dependabotWorkflow` / `dependabotWorkflowPath` bindings and its two local assert helpers are removed. `dependabot-automerge.yml` stays in `LINE_MATCHED_FILES`: nothing line-matches it now, but the LF pin is a real guarantee and cheap to keep — the comment there says so.
- `docs/specs/release.md` — `[GITHUB-DEPENDABOT]` claimed security bumps were gated by dependency review *alone*, which under-describes what the workflows do post-#392. The property is promoted to its own `[GITHUB-DEPENDABOT-SECURITY-GATES]` section naming all four gates, the read-only-token constraint, and the inversion argument.
- `Makefile` — `deployment-verify` gains `_repo-script-deps` as a prerequisite and one line invoking the new suite. `_repo-script-deps` installs the repo-root devDependencies, keyed on npm's own `node_modules/.package-lock.json` so a lockfile bump re-installs rather than running against a stale parser.

**New dependency**

`package.json` / `package-lock.json` — a private repo-root manifest whose only devDependency is `yaml`. `scripts/actions/action-yaml.mjs` hand-scans YAML today specifically because no parser sat on that dependency path, but that scanner is text pattern matching over structured data. Adding the parser closes this properly and makes deleting the scanner possible later. CI already sets up Node 22.x at `.github/workflows/ci.yml:295` in the same `repository-tests` job that runs `make deployment-verify` at `:323`, so no workflow change is needed.

**No breaking changes.** No public API, CLI flag, or report format changes. `[GITHUB-DEPENDABOT]` keeps its id; `[GITHUB-DEPENDABOT-SECURITY-GATES]` is added alongside it.

The suite is black-box: it drives the real workflow files and asserts against a parse of them, reaching into no internals.

Each of its five tests was verified by reintroducing the exact defect it exists to catch and watching it fail for the real reason — 11 mutants, 11 killed:

| Mutant | Killed by |
|---|---|
| `if:` skip on `ci.yml` `changes` (the classifier 9 of 11 jobs key off) | `securityBumpAgainstMainClearsEveryGate` |
| `if:` skip on `ci.yml` `security` | `securityBumpAgainstMainClearsEveryGate` |
| `if:` skip on `codeql.yml` `analyze` | `securityBumpAgainstMainClearsEveryGate` |
| `if:` skip on `action-selftest.yml` `contract` | `securityBumpAgainstMainClearsEveryGate` |
| `env: ACTOR: ${{ github.event.pull_request.user.login }}` rename-evasion | `securityBumpAgainstMainClearsEveryGate` |
| CodeQL `security-events` downgraded to `read` | `securityBumpGatesRaiseTheTokenAboveDependabotReadOnly` |
| dependency review `pull-requests` downgraded to `read` | `securityBumpGatesRaiseTheTokenAboveDependabotReadOnly` |
| an ecosystem drops `target-branch` | `everyEcosystemTargetsTheStagingBranch` |
| sweep subscribes to `main` *(moved assertion)* | `theSweepIsTheOnlyActorGatedJobAndStagingOnly` |
| sweep drops the `head_ref` gate *(moved assertion)* | `theSweepIsTheOnlyActorGatedJobAndStagingOnly` |
| sweep switches to `pull_request_target` *(moved assertion)* | `theSweepIsTheOnlyActorGatedJobAndStagingOnly` |

Two things that table is doing beyond "tests pass":

1. **All five gate mutants survive `main` as it stands today.** The existing `test-release-workflow-contract.mjs` stays green through every one of them. That is the regression hole #394 was filed for, reproduced rather than argued.
2. **The three *(moved assertion)* rows** are the evidence that relocating `dependabotSweepLeavesNoDeadCheckOnHumanPullRequests` preserved its assertiveness. Each corresponds to one assertion of the deleted test, now enforced from its new home against a parse instead of a substring search.

The suite also fails loudly rather than vacuously when misplaced: run from the wrong directory depth every test raises `ENOENT` instead of passing on `undefined`, and `pullRequestBases` rejects a non-array `on.pull_request.branches` rather than treating a parser change as "no branches configured".

Coverage thresholds in `coverage-thresholds.json` are unchanged — this adds Node contract tests, not Rust code.

## For AI

Defect class: **false-negative gate evaluation via workflow-level short-circuit**, not a detector-pipeline defect. No `crates/` code is touched; the accuracy surface here is CI's own gating, where a skipped required check is indistinguishable from a passing one at the branch-protection layer.

Assertion substrate is `yaml.parse` → plain JS object graph. `expressionsIn()` is a generator doing a depth-first walk of the parsed tree, yielding `{path, expression}` for every bare `if:` scalar and every `${{ … }}` interpolation found in any string node; `interpolations()` splits on the literal `${{` / `}}` delimiters rather than matching a pattern, per the repo-wide prohibition on regex over source and structured data. `authorReferences()` filters that stream against `AUTHOR_CONTEXTS`. Consequence: detection is context-name-based and depth-independent, so it catches a skip relocated from `jobs.<id>.if` into `jobs.<id>.env.<KEY>` (the #388 shape) or into a step-level conditional, and it is invariant under YAML reflow, anchor use, and comment churn.

`GATES` and `REQUIRED_WRITE_SCOPES` are declarative tables; adding a fifth gate is a row, not a function. `gatesByWorkflow()` groups by workflow so a single `ci.yml` short-circuit reports both gates it costs rather than one.

Token-scope assertions encode a GitHub-specific invariant worth restating: a workflow run triggered by `dependabot[bot]` receives a read-only `GITHUB_TOKEN` regardless of repository default, so `pull-requests: write` and `security-events: write` at job level are load-bearing for exactly this PR class and inert for every other. Deleting either breaks the gate only on Dependabot PRs — the failure mode is invisible to human-PR testing.

`withFound()` in `contract-assertions.mjs` branches on `Array.isArray` specifically to avoid an error-message regression: the two consumers pass different haystack types (parsed branch list vs. raw workflow text), and unconditional `JSON.stringify` would inline a multi-KB file into the failure of every text assertion in the older suite.

Known pre-existing conditions this PR deliberately does not expand into: `assertIncludes` / `assertEqual` are independently redefined across `scripts/deployment/*.mjs`, `scripts/release/test-release-version-stamping.mjs` and `scripts/benchmarks/*.mjs` (the version in `test-release-version-stamping.mjs` has 2-arity, so it is not a drop-in); and `.devcontainer/devcontainer.json` declares no Node feature despite `typediagram-gen` and eight existing `node scripts/…` invocations already requiring one — `npm ci` adds no tool requirement beyond the Node that was already assumed.

Works on #394

Related: #388 is already fixed by #392 (skips present at `f92300e5`, absent at `2dae2a5b`) and can be closed. #419 is superseded — its workflow edits re-delete those same skips, which is why it conflicts across six files; only its test was worth keeping and it is here, rebased onto the restructured `scripts/` layout.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 # human entry points and are the only ones `make help` lists.
 # =============================================================================
 
-.PHONY: build dup-gate test test-ollama lint fmt clean ci ci-ollama setup help deployment-verify compile-release-tests coverage coverage-run coverage-report _ci-analyze _ci-contract-tests _ci-build _ci-gate _ci-test _ci-test-rust _ci-test-vsix vsix-package vsix-rebuild android-studio-rebuild android-studio-rebuild-reinstall typediagram-gen _delete-path-binaries _kill-deslop-processes _vsix-install _vsix-build _vsix-test _vsix-test-ollama _vsix-coverage _vsix-webview-coverage _vsix-webview-coverage-check _vsix-playwright-html _vsix-install-code _vsix-clean _vsix-stage-bundled-binaries _vsix-stage-and-package _jetbrains-build _jetbrains-verify _jetbrains-package _jetbrains-test _jetbrains-real-binary-test _android-studio-install _android-studio-uninstall
+.PHONY: build dup-gate test test-ollama lint fmt clean ci ci-ollama setup help deployment-verify _repo-script-deps compile-release-tests coverage coverage-run coverage-report _ci-analyze _ci-contract-tests _ci-build _ci-gate _ci-test _ci-test-rust _ci-test-vsix vsix-package vsix-rebuild android-studio-rebuild android-studio-rebuild-reinstall typediagram-gen _delete-path-binaries _kill-deslop-processes _vsix-install _vsix-build _vsix-test _vsix-test-ollama _vsix-coverage _vsix-webview-coverage _vsix-webview-coverage-check _vsix-playwright-html _vsix-install-code _vsix-clean _vsix-stage-bundled-binaries _vsix-stage-and-package _jetbrains-build _jetbrains-verify _jetbrains-package _jetbrains-test _jetbrains-real-binary-test _android-studio-install _android-studio-uninstall
 
 _JETBRAINS_DIR := clients/jetbrains
 
@@ -420,16 +420,30 @@ dup-gate: build
 ##                    action's own step body against the freshly built CLI in
 ##                    both gate directions — the self-test's runner leg cannot,
 ##                    since it installs a published release ([ACTION-GATE]).
-deployment-verify: build
+deployment-verify: build _repo-script-deps
 	node scripts/deployment/verify-deployment-manifest.mjs shipwright.json
 	node scripts/deployment/verify-deployment-binaries.mjs shipwright.json target/release
 	node scripts/release/verify-release-workflow-gates.mjs .github/workflows/release.yml
 	node scripts/release/test-release-workflow-contract.mjs
+	node scripts/release/test-dependabot-gate-matrix.mjs
 	node scripts/deployment/test-deployment-docs-contract.mjs
 	node scripts/release/test-release-version-stamping.mjs
 	node scripts/deployment/test-verifiers.mjs
 	node scripts/actions/test-action-contract.mjs
 	node scripts/actions/test-action-diff-gate.mjs
+
+# _repo-script-deps: Install the repo-root devDependencies the Node contract
+#   suites under scripts/ import (`yaml`, so workflow gates are asserted against
+#   a real parse, never a text match). Root package.json is private and ships
+#   nothing — the VSIX remains the only distribution.
+#   Keyed on npm's own `node_modules/.package-lock.json`, which it rewrites on
+#   every install: a mere `-d node_modules` test would skip the install after a
+#   lockfile bump and run the suite against the stale parser.
+_repo-script-deps:
+	@if [ ! -f node_modules/.package-lock.json ] \
+	   || [ package-lock.json -nt node_modules/.package-lock.json ]; then \
+	  echo "==> Installing repo-root script dependencies..."; npm ci; \
+	fi
 
 # _kill-deslop-processes: SIGTERM (then SIGKILL on holdouts) every running
 #   `deslop-lsp` and `deslop-mcp` process so a stale child from a previous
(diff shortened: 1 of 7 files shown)


MelbourneDeveloper:
Sorry, no time to review these ones. We are working on accuracy issues at the moment.

Issue #439, opened 2026-08-23
type2_recall is blind: the only curated Type-2 recall check passes with the curated duplicate deleted


abdushakoor12:
## The defect

[`crates/deslop-test-support/src/corpus_confidence.rs:357`](https://github.com/Nimblesite/Deslop/blob/main/crates/deslop-test-support/src/corpus_confidence.rs#L357) — `check_one_curated_type2` — is the only assertion guarding Type-2 rename recall on real-repository ground truth. Both of its clauses ask the same weak question: *does **some** cluster touch these two paths?*

```rust
if !reports_clone_spanning(report, &files) {          // :360 — any cluster, any extent
    …"no cluster spans {files:?}"…
}
let vouched = visible_clusters(report)                // :367
    .iter()
    .any(|cluster| gate_vouched(cluster) && cluster_shows_span(cluster, &files));  // :369
```

`reports_clone_spanning` ([`corpus.rs:541`](https://github.com/Nimblesite/Deslop/blob/main/crates/deslop-test-support/src/corpus.rs#L541)) and `cluster_shows_span` ([`corpus.rs:591`](https://github.com/Nimblesite/Deslop/blob/main/crates/deslop-test-support/src/corpus.rs#L591)) are pure path-set membership. Neither reads the cluster's extent, node count, size or rank. So the check cannot distinguish **"reported as one module"** from **"reported as a buried fragment"** from **"not reported at all, but the two files happen to share a five-line accessor"** — the three outcomes its own doc comment names as the failures it exists to catch.

`corpus/tokio.json` curates the pair as *"the two standard-stream handles are one module written twice"* — 134 and 183 lines, every declaration and impl header renamed. Both witnesses below satisfy the check while that claim is false.

## Witness 1 — the check passes with the curated duplicate deleted

Scan of the pinned `tokio-1.49.0` clone at `7bb29d4`, with the curated module cluster removed from the rendered report and both clauses re-applied:

```
clusters after deleting the curated module pair: 2396 of 2397
clause 1 reports_clone_spanning: True
clause 2 shown gate-vouched    : True
  -> satisfied by rank 56 nearly_identical size=17 nodes=31 rc=0.62
     unrelated files it also spans: 11, e.g. tokio/src/io/stdin.rs,
     tokio/src/net/tcp/listener.rs, tokio/src/net/tcp/stream.rs
```

[...]

Issue #431, opened 2026-08-22
Content gate fabricates token_jaccard = 1.0 for clusters that are not Merkle-identical (STRUCTURAL_SATURATION_FLOOR = 0.99)


MelbourneDeveloper:
## Where

`crates/deslop-core/src/buckets/gate.rs` on `main` (`42b2c92`), lines 204-210, inside `apply_content_gate`:

```rust
let token_jaccard = if kind == ClusterKind::NearlyIdentical
    && signals.structural >= STRUCTURAL_SATURATION_FLOOR
{
    1.0
} else {
    signals.token_jaccard
};
```

with `pub const STRUCTURAL_SATURATION_FLOOR: f64 = 0.99;` at line 78.

## The defect

The comment above the correction states the justification:

> A shape-identical cluster routed `NearlyIdentical` shares one Merkle hash, so the members' normalised kind streams are equal by construction and the true token Jaccard is 1.0 [...] The `structural` guard scopes the correction to clusters the Merkle argument actually covers.

The Merkle argument holds at `structural == 1.0` and nowhere else. Since #408 measures `structural` as **subtree overlap** — a ratio, not a digest equality flag — any value strictly below 1.0 means the members' subtrees provably differ, so there is no shared Merkle hash and the kind streams are *not* equal by construction.

The guard nevertheless admits the whole band `[0.99, 1.0)`. For every cluster in that band the correction does not repair a fallback-signature artifact; it overwrites a measured signal with a value no evidence supports. The rendered `token_jaccard = 1.0000` then asserts token agreement that was never observed, and because `fused` is computed from `shape_score()` and the content confidence, the inflated pair is pushed toward the act-now line.

## Reproduction

Two Python files differing in exactly one operator token (`+` -> `-`) — they compute different answers:

```python
# ledger_credit.py
def post_credit(entries, rate, floor, ceiling):
    running = 0
    adjusted = []
    for entry in entries:
        scaled = entry * rate
        shifted = scaled + floor          # ledger_debit.py: scaled - floor
        clamped = min(shifted, ceiling)
        adjusted.append(clamped)
        running = running + clamped
    average = running / max(len(adjusted), 1)
    spread = max(adjusted) - min(adjusted)
    summary = {"total": running, "average": average, "spread": spread}
    return summary
```

Measured cluster:

```
id=3c351ea8ee5cb48d  bucket=nearly_identical  category=logic  size=2  weight=101.400
[...]

Issue #427, opened 2026-08-21
JavaScript reports the nested fragment where TypeScript reports the enclosing method pair (#408 not closed for JS)


MelbourneDeveloper:
The `javascript-type3` and `typescript-type3` fixtures hold the *same* source shape — `accumulate`/`aggregate` with a full identifier rename plus one extra statement (`running = running + 2;`) in `delta`. The two languages elect different views.

Measured at `--min-nodes 8 --embeddings off`:

```
javascript-type3   visible 1  hidden 0
  nearly_identical  structural 1.0     fused 0.9474   delta.js 2-9   epsilon.js 2-9
typescript-type3   visible 1  hidden 1
  nearly_identical  structural 0.8542  fused 0.7857   delta.ts 1-12  epsilon.ts 1-11
```

TypeScript publishes the enclosing function pair. JavaScript publishes only the nested `let running = 0; for (...) { ... }` run and never reports the method-level clone at all — the enclosing pair is not even hidden, it is absent.

This is the defect [FUSION-SHARED-SUBTREE] describes and claims to have closed: "A single inserted statement rehashes every ancestor Merkle node, so a textbook Type-3 near-miss — full identifier rename plus one extra statement — scored `structural = 0.0` on the enclosing method while the unchanged statements *inside* it stayed Merkle-identical. The fragments were reported and the method was not, in **every** language." JavaScript still does the old thing.

A user reading the JavaScript report is told two loop bodies are similar, when what they should act on is that two whole exported functions are near-copies.

Not pinned today: `crates/deslop/tests/type3_enclosing_method.rs` covers `csharp-type3`, `dart-type3`, `go-type3`, `python-type3` and `ts-type3-stmt`, but not JavaScript. `crates/deslop/tests/js_ts_signatures.rs::javascript_near_miss_produces_cross_file_structural_cluster` passes precisely *because* the fragment is what gets published — its `structural == 1.0` assertion can only hold for a Merkle-equal view.

Repro:

```
deslop crates/deslop/tests/fixtures/javascript-type3 --min-nodes 8 --embeddings off
deslop crates/deslop/tests/fixtures/typescript-type3 --min-nodes 8 --embeddings off
```

Expected: JavaScript reports `delta.js` 1-12 against `epsilon.js` 1-11, matching TypeScript. A `javascript-type3` row belongs in `type3_enclosing_method.rs` once it does.

[...]

Issue #408, opened 2026-08-18
STOP-class: spec and code promise Type-3 renamed near-miss recall that does not exist in any language


MelbourneDeveloper:
## The disagreement

`crates/deslop-core/src/buckets.rs:270-279` and `docs/specs/fusion.md` both state that the renamed Type-3
near-miss is protected, and name `csharp-type3` as the measured case:

> Measured against a genuine Type-3 clone whose identifiers are all renamed and whose bodies differ by one
> statement (`csharp-type3`) ... Gating row 4 here therefore demotes the renamed near-miss, **the most
> valuable clone class there is**. Row 4 is instead routed on cluster *spread* in
> `report_render::route_shape_identical`.

The promise is that row-4 spread routing keeps this clone class visible. **Measured, it does not.**

## The fixture

`crates/deslop/tests/fixtures/csharp-type3/` — a full identifier rename (`Delta→Epsilon`,
`Service→Worker`, `Aggregate→Process`, `bound→limit`, `running→accumulator`, `step→cursor`) plus exactly one
inserted statement inside the loop:

```csharp
// Delta.cs                                  // Epsilon.cs
public int Aggregate(int bound)              public int Process(int limit)
{                                            {
    if (bound < 0) { return 0; }                 if (limit < 0) { return 0; }
    int running = 0;                             int accumulator = 0;
    for (int step = 0; step < bound; step++)     for (int cursor = 0; cursor < limit; cursor++)
    {                                            {
        running = running + step;                    accumulator = accumulator + cursor;
        running = running + 2;   // <-- only  }
    }                            //     diff   return accumulator;
    return running;                          }
}
```

Any developer would call these duplicated. This is the textbook Type-3 clone.

## Measured

```
deslop <csharp-type3> --min-nodes 8   ->  clusters=2 hidden=0
   structural_only  fused=0.667  nodes=13  [Delta.cs:7,  Epsilon.cs:7]
   structural_only  fused=0.573  nodes=13  [Delta.cs:12, Epsilon.cs:12]
deslop <csharp-type3> --min-nodes 12  ->  identical result
```

**The whole-method pair never forms a cluster.** The only things reported are two 13-node *fragments* — the
`if (bound < 0) { return 0; }` guard and the loop header — and both are demoted to `structural_only`, whose
rendered text reads "Same shape, different content ... usually sibling boilerplate." That is false about this
pair on both clauses.

## It is every language, not just C#

[...]

PR #450, opened 2026-08-25
Stop the content gate publishing an unmeasured token_jaccard (gh #431); pin the JS/TS enclosing-view election (gh #427)

abdushakoor12:
## TLDR

Stops the content gate publishing a `token_jaccard` it never measured (gh #431), and pins the enclosing-view election in JavaScript and TypeScript so the Type-3 contract covers all seven languages instead of five (gh #427).

## Details

Two rendered-accuracy defects, both exposed by the `structural` axis becoming a *graded* subtree overlap in #408.

**gh #431 — the content gate fabricated two published signals.**
`apply_content_gate` rewrote `token_jaccard` to `1.0` for `nearly_identical` clusters on a Merkle argument: members sharing one normalised-subtree digest have equal kind streams by construction, so a lower measured value is a fingerprint-scoped fallback-signature artifact rather than evidence. That argument is sound — but it holds at digest equality and nowhere else. The guard it was given was `structural >= STRUCTURAL_SATURATION_FLOOR` (0.99), a near-miss **routing** tolerance. Since `structural` grades overlap, every value in `[0.99, 1.0)` reached that guard, and every one of them means the members' subtrees provably differ.

Two published numbers were fabricated for every cluster in that band: `token_jaccard` directly, and `shape` — `max(structural, token_jaccard)` — which inherits it. `fused` is computed from the pre-correction triple, so it was unaffected; that split is what makes the defect observable from the report alone, because the rendered signals stopped reproducing the rendered confidence. Measured on the operator-drift pair (one `+`/`-` apart inside a twelve-line shared body): `structural` 0.9907, published `token_jaccard` 1.0000 against a measured 0.9141, published `shape` 1.0000. The rendered human verdict narrated it as "The shapes match at 1.00".

The guard is now the digest equality the argument actually names. `report_render.rs` gains `members_share_one_digest`, which tests the fact on the members, and `buckets::content_gated_signals` / `apply_content_gate` take it as an argument. No `structural` reading can substitute: `measure_views` returns `shared / larger`, so saturation is reachable by ratio without hash equality. **`fused`, bucket routing, weight and rank are unchanged by construction** — this corrects rendered figures only. Spec: `[FUSION-CONTENT-GATE]` step 5 in [`docs/specs/fusion.md`](docs/specs/fusion.md) now scopes the correction to digest equality and explains why no `structural` reading can stand in for it.

**gh #427 — the JS/TS enclosing-view election was never covered end to end.**
The engine fix landed with the `[PIPELINE-CLUSTER-EXACT-SCOPE]` guard, but nothing tested it on JavaScript or TypeScript, and two assertions elsewhere still described the tree it left behind. `javascript-type3` and `typescript-type3` hold the same source shape as `ts-type3-stmt` — `accumulate`/`aggregate`, a full identifier rename plus one trailing `running = running + 2;` — and both reported the nested `let running = 0; for (…)` run at lines 2–9, dropping the method pair entirely.

Admission was never the cause, which is what separates #427 from the five languages already pinned: the rescue measures the enclosing pair at `overlap = 0.8679`, clearing both shared-subtree floors, so it enters clustering. It was lost in the same-file overlap collapse, where the Merkle-equal fragment reads 1.00 *because* it excludes the inserted statement. TypeScript reached the guard first only because its type annotations put a fingerprint boundary strictly inside the declaration — an accident of that grammar, not a rule.

`type3_enclosing_method.rs` gains `javascript-type3` and `typescript-type3` rows on the existing five-language contract, so one contract now holds all seven. `[PIPELINE-CLUSTER-EXACT-SCOPE]` in [`docs/specs/pipeline.md`](docs/specs/pipeline.md) is restated: the rule protects an encloser that sits inside the same declaration **or inside none at all**, because the view a Type-3 near-miss wants published is frequently the declaration itself, for which the scope resolver reports no enclosing declaration by design.

**Also in this diff.** `polymorphic_gate_hides_rename_clone.rs` moves 2 hidden groups to 3 — the extra group previously collapsed below `MIN_REPORTABLE_MEMBERS` and now materialises and is suppressed by the polymorphic filter — with a new `Found 0 groups` assertion added alongside so the count is anchored to *why* it is 3 rather than free to drift. Three curated-recall pins for gh #439 (`type2_recall` compares curated paths but not curated extent) are declared `#[ignore = "[SKIP-UNFINISHED] GH #439 …"]` under `[TEST-SELECTION-SKIP]`, registered in `skip_policy_contract.rs`, and documented as § L9 in [`docs/plans/corpus-assertion.md`](docs/plans/corpus-assertion.md). No assertion was deleted or weakened; each returns when the extent predicate lands.

No breaking changes: no public API, CLI flag, report field or spec id was removed or renamed.

**`crates/deslop/tests/content_gate_signal_honesty.rs`** (new, 2 tests, both watched red before the fix):

- `the_content_gate_publishes_no_token_jaccard_it_did_not_measure` drives the CLI over the operator-drift fixture and asserts against the rendered report that the cluster in the `[0.99, 1.0)` structural band publishes its **measured** `token_jaccard` (0.9141), not 1.0, and that the derived `shape` no longer inherits the fabricated value. It fails on the old guard for the right reason: the value it reads back is exactly the one the Merkle argument does not cover.
- `a_digest_equal_pair_keeps_its_saturated_signals` is the other direction, and is what stops the fix being a blanket removal: a genuinely digest-equal `nearly_identical` pair still gets the correction to 1.0. Without this, deleting the correction outright would pass the first test.

**`crates/deslop/tests/type3_enclosing_method.rs`** — the five-language contract becomes seven. `javascript_type3_reports_the_enclosing_method_pair` and `typescript_type3_reports_the_enclosing_method_pair` assert, per language, the **exact occurrence spans in both files** (`delta.js` 1–12 / `epsilon.js` 1–11 and the `.ts` equivalents), the cluster size, the bucket, the fused floor, and that **no fragment view is visible beside the pair** — that last one is what catches the regression, since the old behaviour published the nested run *instead of* the pair. Both were watched red before the engine fix and are green after it.

**Regression surface.** Reverting `members_share_one_digest` to the `structural` floor fails `the_content_gate_publishes_no_token_jaccard_it_did_not_measure`; reverting the scope guard fails both new Type-3 rows. `skip_policy_contract.rs` (7 tests) holds the three new `#[ignore]`s to a category, an issue, a spec id, a plan document that names that issue, and a run instruction — a bare or undocumented skip fails the gate, and `the_ignored_tests_in_the_tree_are_exactly_the_curated_set` fails if a skip is added or removed without curating it.

Coverage thresholds in `coverage-thresholds.json` are unchanged by this PR; `make ci` passes locally with coverage at or above every floor.

## For AI

Two independent renderer-layer accuracy defects, both downstream of `[FUSION-SHARED-SUBTREE]` regrading `signals.structural` from digest-equality to graded subtree overlap (`shared / larger` from `measure_views`).

`#431`: `buckets::gate::apply_content_gate` conditioned the Merkle token-axis correction on `signals.structural >= STRUCTURAL_SATURATION_FLOOR` (0.99). Post-#408 that predicate is satisfiable by ratio, decoupling it from the `hash`-equality invariant the correction's soundness proof requires. Repair threads a `members_share_one_digest: bool` from `report_render::cluster_to_report` through `content_gated_signals` → `apply_content_gate`, computed by folding `Fingerprint::hash` equality across `cluster.members` (empty ⇒ `false` via `Option::is_some_and`, so a memberless cluster asserts nothing). `fused` derives from the pre-correction triple, so routing/weight/rank are provably invariant; only `token_jaccard` and the derived `shape = max(structural, token_jaccard)` move. Observability of the defect from the wire report follows from that same asymmetry.

`#427`: `[PIPELINE-CLUSTER-EXACT-SCOPE]` in `Occurrence::same_declaration_scope` — the `(None, Some(_)) => true` arm is the load-bearing one, admitting a candidate whose own declaration scope is `None` (a view that *is* the declaration, or the `export` wrapper the grammar hangs above it) as an eligible encloser. The prior symmetric-strict formulation left whole-declaration views unprotected against Merkle-equal nested windows, which win the same-file overlap collapse's cross-file-edge contest precisely by excluding the discriminating statement. TS masked it: type annotations induce a fingerprint boundary strictly interior to the declaration, so its representative carried `Some(decl)` incidentally. Non-comparability of grades across nesting scopes is the same invariant `[PIPELINE-CLUSTER-SUBSUME]` enforces one stage earlier; no scalar threshold separates the cases (encloser must win at 0.857/0.868, lose at 0.879 in `fsharp_issue_339_sibling_window_rename`; node-mass and byte-share invert with it).

`#439` pins are quarantined via the `[TEST-SELECTION-SKIP]` registry rather than deleted, preserving the assertions as an executable specification of the missing extent predicate (`min_nodes` vs `canonical_node_count` in `check_one_curated_type2`). Run them with `-- --ignored`.

Refs gh #431, gh #427, gh #408, gh #439.


MelbourneDeveloper:
Reviewed against `origin/main` (branch is 0 behind, merge base = `0d8f82f`).

**The invariance claim holds** — I traced it rather than taking the body's word: the bucket label comes from `raw_signals` before the gate, the hide decision from the wire label plus raw `cluster.signals` (`report/hidden.rs:102`), and weight from `signals.fused` plus the wire label. `fused` is unchanged, confirmed arithmetically against #431's own cluster: `0.9907 × 0.9565 = 0.9476` vs the `0.9477` it reports. This really is render-only.

**One real gap.** `a_digest_equal_pair_keeps_its_saturated_signals` asserts `bucket == "identical"`, but `content_gated_signals` early-returns on `ClusterKind::Identical` and never reaches `apply_content_gate` — the control's saturation comes from `proven_identical_signals`. So hardcoding `members_share_one_digest` to `false` (i.e. deleting the correction outright) passes **both** new tests, which is exactly what the body says this test prevents. It needs a `nearly_identical` digest-equal pair — a consistent rename is Merkle-equal without being byte-equal.

**The polymorphic 2→3 can't come from this diff.** `MIN_REPORTABLE_MEMBERS` is upstream in `cluster.rs` and the gate change provably can't affect visibility. Branch 0 behind main, main asserts 2, CI green at 3 → `hidden_group_summary_names_the_hider_not_the_users_config` is almost certainly already red on `main`, likely from `c3ce7882` (pushed without a PR gate). Worth saying so in the body rather than filing it under "also in this diff".

No assertions removed anywhere in the patch; the three `#[ignore]`s are new red pins for #439, not weakening. Nits: stray blank line between `#[test]` and `fn` in the two new `type3_enclosing_method.rs` rows, and the comment at `pipeline/signatures/tests/fold_parity.rs:138` still describes the old guard.

Issue #416, opened 2026-08-18
dependency_reactivity.rs asserts on unwrap_or_default(): a missing count field reads as 0 and the < 4 bound passes vacuously


MelbourneDeveloper:
## Defect (on main)

\`crates/deslop-lsp/tests/dependency_reactivity.rs\` (main, ~line 108–120): the assertion reads a count out of the notification JSON as

    .as_u64().unwrap_or_default() < 4

If the field is renamed, absent, or the notification shape changes, \`unwrap_or_default()\` yields \`0\`, \`0 < 4\` holds, and the test passes without having observed the behaviour it exists to bound. Fail-open: the reactivity regression it guards could return undetected.

## Fix shape

A missing field must fail the test, not default: \`ok_or\`/\`expect\` with a message naming the field, and assert the positive expected value (or exact range) rather than only an upper bound against a defaulted zero.

Found during the #364 branch audit; verified present on \`origin/main\` (\`unwrap_or_default\` at line 114).

MelbourneDeveloper:
Related: #415 — the same fail-open assertion pattern in the CLI fused-score bounds suite.

MelbourneDeveloper:
**Cluster: the release gate asserts nothing** — #412 #414 #415 #416 #398 #401. Root cause: #412 — fix it first; the rest may collapse or shrink once it lands. Fix order: #412 → #414 → #401 → #415 → #416 → #398.

Every member is a place where a green gate proves nothing. Verified against `origin/main` @ `42b2c928c`:

- **#412 (root)** — `Makefile:75` still reads `-- --skip ollama_ --skip corpus_`. cargo's `--skip` is a substring match on the test *name*, not a suite selector, so 30+ tests never execute in `make test` / `make ci`, including the corpus accuracy gate's own self-tests.
- **#414** — what that filter hid: `corpus_flutter_dart` breaches both its own [CORPUS-CEILINGS] bounds (wall 1014.7 s > 900 s, RSS 10177 MB > 7168 MB). Recorded blocked by #412 (the mechanism that hid it) and #166 (the single-threaded / whole-corpus-in-RAM cost the ceiling was bounding).
[...]

Issue #370, opened 2026-08-14
CRITICAL: deslop-lsp hangs indefinitely after a rejected embedding refresh (test ignored, ate the whole CI budget)


MelbourneDeveloper:
## TL;DR

One of our own tests hangs forever instead of failing. It's currently skipped so CI can pass. The hang points at a real product bug: when an embedding refresh is rejected, the language server never reports that it finished or failed — so an editor would sit on a spinner forever with no way to know the refresh died.

## Details

The test starts a real Deslop language server, asks it to refresh embeddings, rejects the request, and waits for a "done or failed" message. That wait has a 20-second timeout and works correctly — the problem is that the timeout is never reached. Before it, the test is stuck on a simpler read with no timeout at all, waiting for a message the server never sends.

Locally it hung for nearly 15 minutes before being killed. On CI it burned the entire test budget twice — once under a 10-minute limit, then again under a raised 30-minute limit. It is now marked `#[ignore]` so the branch can merge; nothing about the test's checks was weakened, and it can still be run on demand.

Why this matters beyond a slow test: if the server genuinely never sends a "finished" or "failed" message after a rejected refresh, then a user's editor waits on a spinner indefinitely. That is exactly the behaviour the test exists to prevent.

It is pre-existing — proven by reverting the branch changes and still hanging past 5 minutes — and had never actually run in CI before, because earlier tests on the branch always died first. Two quick leads for the fix: check whether the server sends any final message at all on the rejection path; and give the test's raw read a timeout, so any future bug of this shape fails fast with a clear error instead of eating CI.

## Steps To Reproduce

Currently skipped — run it explicitly, and expect to kill it manually:

1. Run the test:

   ```
   cargo test -p deslop-lsp --test embedding_failure_progress -- --ignored rejected_embedding_refresh_reports_failure_and_preserves_last_good_report
   ```

2. Observe: no output, no failure — it just hangs (measured 14m41s locally before a manual kill; CI cancelled it at 10m16s and, after the budget was raised, at 30m15s).

## Details (for AI)

```shell
# evidence
test: deslop-lsp::embedding_failure_progress::rejected_embedding_refresh_reports_failure_and_preserves_last_good_report
status: #[ignore]d so the branch can merge; assertion intact (run with -- --ignored)
measured: 14m41s local kill on `fused`; CI cancelled at 10m16s (timeout-minutes: 10), again at 30m15s (budget raised to 30)

# hang mechanics
[...]

Issue #312, opened 2026-07-28
Flaky: refresh_command_re_evaluates_the_corpus_after_an_edit races the live watcher


MelbourneDeveloper:
## TL;DR

One of our own tests passes on a developer machine but fails randomly on CI. The test edits a file and then asks Deslop to refresh its report, expecting the refresh to show the duplicate disappearing. But Deslop's background file watcher often updates the report first — so by the time the test's refresh runs, there is nothing left to report, and the check fails.

## Details

The test writes to a file, then explicitly asks Deslop to refresh the report, then checks that the refresh reported one duplicate removed.

The problem: Deslop also has a background watcher whose whole job is to react to file edits — that is the "live" in deslop.live. On a busy CI machine the watcher usually wins the race: it updates the report, and the removal is recorded there. The test's own refresh then sees an already-up-to-date report with zero changes, and the check for "at least one duplicate removed" fails.

On an idle developer machine the test's refresh usually gets there first, which is why it passes locally. A test whose result depends on who wins a race is exactly what our testing rules forbid ("deterministic — no timing dependencies").

The right fix is to check the final state instead of the race: after the refresh, ask Deslop for the current report and confirm no duplicate spans the two files. That check is both stronger and immune to the race. Do not weaken the assertion and do not add a wait.

This is not caused by the branch it was observed on — that branch changed none of the relevant code. Note the same trap quietly affects any test that edits a file and then checks what an explicit refresh reported.

## Steps To Reproduce

Fails intermittently on busy CI runners; usually passes locally:

1. Run the test:

   ```
   cargo test -p deslop-lsp --test editor_non_interference refresh_command_re_evaluates_the_corpus_after_an_edit
   ```

2. If it passes, run it under load (or on a CI runner) until it fails.
3. The failure shows `clustersRemoved: 0` with `"generation": 3` in the payload — proof the watcher already did the work at generation 2.

## Details (for AI)

```shell
# evidence
test: crates/deslop-lsp/tests/editor_non_interference.rs:232 (file write at :210)
failing run: 30352691142 / job 90253738290 (#311), commit 399084ac0 — passes locally on the identical commit
failure payload: {"clustersAdded":0,"clustersRemoved":0,"clustersUpdated":0,"command":"deslop.lsp.refreshReport","generation":3}

# race mechanics
ReportDelta::between (crates/deslop-core/src/delta.rs:31) diffs next against the immediately preceding snapshot:
[...]

PR #453, opened 2026-08-25
Fail-closed LSP test assertions: exact duplicated-files count (gh #416), race-immune refresh end state (gh #312)

Wamwea:
## TLDR

Two fail-open LSP test assertions can no longer pass without observing anything: the dependency-reactivity duplicate-count read is now fail-closed and exact (gh #416), and the refresh-command test asserts the deterministic end state instead of racing the live watcher (gh #312).

## Details

Test-only change; no production code touched.

**`crates/deslop-lsp/tests/dependency_reactivity.rs` (#416).** The old assertion read `metrics.duplicated_files` as `.as_u64().unwrap_or_default() < 4`, so a report that stopped rendering the field — or a blind detector rendering zero — satisfied the bound vacuously. The read is now a named verdict, `dependency_left_duplicate_population`, comparing the raw wire value against the new constant `DUPLICATED_FILES_AFTER_EDIT = 3`: the measured ground truth that after `node_modules/pkg/Beta.cs` is rewritten to unrelated code, the three untouched copies of the `csharp-small` pair remain one clone family ([PRINCIPLES-LIVE-IS-REACTIVE]). An absent or non-numeric field compares as JSON `Null` and fails loudly. The `"duplicated_files"` literal is hoisted to `DUPLICATED_FILES_FIELD` and reused at both read sites.

**`crates/deslop-lsp/tests/editor_non_interference.rs` (#312).** `refresh_command_re_evaluates_the_corpus_after_an_edit` asserted `clustersRemoved >= 1` on the refresh response — but the delta diffs against whichever snapshot came immediately before, so on a busy runner the live watcher drops the clone first and the refresh legitimately reports zero (the recorded CI failure: `clustersRemoved: 0, generation: 3`). Since `refresh_full()` runs synchronously under the session lock, the end state after the response is deterministic whichever side won. The test now pins the response's `command` echo, then asserts via `deslop/reportGet` that the refreshed report renders zero clusters and both fixture files stay analysed (`FIXTURE_FILE_COUNT = 2`). No sleep added; the generation number is no longer load-bearing. The command literal is hoisted to `REFRESH_REPORT_COMMAND`.

No breaking changes: no public API, CLI flag, report format, or spec ID changed.

Related issues (not auto-closed, per repo policy): gh #416, gh #312.

- **`dependency_reactivity::duplicate_population_verdict_rejects_absent_and_blind_counts` (new)** feeds the verdict the three discriminating reports: `{"metrics": {}}` (field absent) and `{"metrics": {"duplicated_files": 0}}` (blind detector) must be rejected, and `{"metrics": {"duplicated_files": 3}}` must be accepted. It was run red against the old logic first — the absent-field case passed the old bound, which is exactly the #416 defect — and is green only with the fail-closed equality.
- **`dependency_reactivity::opted_in_dependency_edit_refreshes_clusters_metrics_and_occurrences` (strengthened)** drives the real `deslop-lsp` binary over stdio against the seeded four-copy workspace and now requires `duplicated_files` to equal exactly 3 after the dependency edit — a report missing the field, a stale 4, or a blind 0 all fail where the old `< 4` bound passed two of them.
- **`editor_non_interference::refresh_command_re_evaluates_the_corpus_after_an_edit` (de-flaked)** was reproduced red deterministically before the fix by forcing the watcher to win the race (temporary scaffolding, not committed): the old assertion failed with the exact CI payload `clustersRemoved: 0, generation: 3`. The rewritten end-state assertion was then verified green on **both** sides of the race — with the forced watcher win and 5/5 consecutive runs on the fast path — and cannot pass blind, because the same test's pre-edit phase (`wait_for_clusters`) requires the clone pair to be detected first.
- Full `deslop-lsp` suite in the gate profile (`cargo test --release --features deslop-lsp/profiling --test suite`): 71 passed, 0 failed, 5 ignored (the tracked `[TEST-SELECTION-SKIP]` registry).

## For AI

Fail-open instrument repair; production honest, tests were not falsifiable. #416: `serde_json` coercion chain `Value -> Option<u64> -> unwrap_or_default()` maps {absent, non-numeric} -> 0, and `0 < 4` is a tautology over that image; replacement compares `&Value == u64` via serde_json's integer-semantic `PartialEq`, so `Null` and `0` are both outside the accepted singleton {3}. Ground truth 3 derives from the fixture topology (2x2 seeded copies of `csharp-small`, one member rewritten) and was validated against the live engine, not calibrated to it. #312: `ReportDelta::between` diffs adjacent generations; watcher-debounce vs explicit `refreshReport` is a benign race for state but not for deltas — the delta observable is generation-order-dependent, the post-refresh report observable is not, because `refresh_full()` is synchronous under the session mutex and the corpus is quiescent after the single write. Assertion migrated from the racy observable to the invariant one per the issue's acceptance criteria; scaffolding used to force the loser's schedule during red-phase verification was deleted after capture. Related open work not addressed here: #352 (`cpu_throttle_knob` 250 ms sleep probe), #370 (unbounded `recv_response`), which share the harness but not this defect class.


MelbourneDeveloper:
Please run /ci-prep before submitting

image

